PR #10, opened 2025-09-04
Change ret type of binding constructor to interface

RodrigoAD:
### What
- Constructing contract bindings return an Interface instead of the contract struct

### Why
Gives more flexibility on using and testing bindings


JoaoVasques (approved):
lgtm 👍

PR #18, opened 2025-09-09
add an upgrade registry and tests

FelixFan1992:


contracts/ccip/ccip/sources/upgrade_registry.move
```diff
@@ -0,0 +1,132 @@
...
+    version: u8,
+}
+
+const EFunctionNotAllowed: u64 = 1;
+
+public struct UpgradeRegistry has key, store {
+    id: UID,
+    // module_name -> vector[vector<u8>, vector<u8>, ...]
+    // the module_name represents the module under which the function is located, e.g. "fee_quoter", "offramp", etc.
+    // the outer vector includes all the blocked function versions for this given module_name
```
JoaoVasques, line 26: Just for clarification, does the position of an inner vector in the outer vector represents all the restrictions of a version? Sharing an example below 

```json
{
  "fee_quoter": 
     [0, "foo"],
     [1, "bar"]
}
```

In this example, if we want to access the restrictions of version 0 of `fee_quoter` we could do it in `O(1)` time by `registry["fee_quoter"][0]`

FelixFan1992: i don't think it currently works this way. we would not be able to get all the functions under a specific versions in O(1).

bc the inner vector contains vectors, so traversing will take O(n).
e.g.
your example will be
```
{
  "fee_quoter": 
     [0, 102, 111, 111],
     [1, 98, 97, 114],
}
```

these function names are appended to the version number and they become an identifier for this function at this version.

if all the functions at a version is blocked, it will be a vector of a single version number `[2]`. but nevertheless, the inner vector takes O(n) to search.

JoaoVasques: ah ok got. Not a big problem, was just for clarification

faisal-chainlink (approved):

JohnChangUK (approved):

JoaoVasques (approved):
Great job

PR #35, opened 2025-09-12
More cs

stackman27:


FelixFan1992 (approved):

deployment/state.go
```diff
@@ -11,19 +11,21 @@ type CCIPChainState struct {
 	CCIPRouterAddress          string
 	CCIPAddress                string
 	CCIPObjectRef              string
+	CCIPOwnerCapObjectId       string
 	MCMsAddress                string
 	TokenPoolAddress           string
 	LockReleaseAddress         string
-	LockReleaseStateID         string
-	FeeQuoterCapID             string
+	LockReleaseStateId         string
```
FelixFan1992, line 18: why we change these to Id but change those in state/type.go to ID?

stackman27: i think go lint prefers "ID" so best to change it to ID whereever we can

PR #36, opened 2025-09-14
Hexify return bytes from CR

stackman27:


relayer/common/utils.go
```diff
@@ -52,3 +53,27 @@ func SerializeUBigInt(size uint, v *big.Int) []byte {
 
 	return ub
 }
+
+// ConvertBytesToHex recursively walks through any value and hex-encodes all []byte values.
+func ConvertBytesToHex(value any) any {
```
FelixFan1992, line 58: can we add some test cases for this?

faisal-chainlink: Re- tests I think it should be in the `loop_reader_test.go` file instead of just a unit test. I'm working on a separate PR that improves the loop reader tests.

faisal-chainlink (approved):

FelixFan1992 (approved):

PR #40, opened 2025-09-15
simplify rmn remote and offramp

FelixFan1992:
there will be no more `blessed_merkle_roots` from rmn. hence, removing related logic. but keeping the event sig the same for compatibility for offchain and o11y


contracts/ccip/ccip_offramp/sources/offramp.move
```diff
@@ -983,14 +983,15 @@ public fun commit(
...
 
-    if (commit_report.blessed_merkle_roots.length() > 0) {
-        verify_blessed_roots(
-            ref,
-            object::uid_to_address(&state.id),
-            &commit_report.blessed_merkle_roots,
-            commit_report.rmn_signatures,
-        );
-    };
+    // // there will be no blessed merkle roots
```
JohnChangUK, line 986: Do we need this commented?

JohnChangUK (approved):

PR #41, opened 2025-09-15
Feature: Rename response fields from CR functions

faisal-chainlink:


relayer/chainreader/reader/chainreader.go
```diff
@@ -195,6 +195,11 @@ func (s *suiChainReader) GetLatestValue(ctx context.Context, readIdentifier stri
 			structResult[mapKey] = results[i]
 		}
 
+		// Apply result field renames if configured
+		if functionConfig.ResultFieldRenames != nil {
+			structResult = applyResultFieldRenames(structResult, functionConfig.ResultFieldRenames).(map[string]any)
```
stackman27: same comment as (https://github.com/smartcontractkit/chainlink-sui/pull/42) 

why not keep this outside the if statement, so we always apply renames if `ResultFieldRenames` is present, we won't use it in Core and only be used in o11y, so we should be safe right?

faisal-chainlink: What do you think about keeping it since we don't lose anything by having it as a feature for non-o11y workflows? the cost here is just an `if` statement since we're returning and exiting. 

I do agree that the refactoring could make this entire method better and as mentioned I think we will benefit from doing that as a separate PR and refactoring some bigger chunks of the code to use bindings for reads.

FelixFan1992 (dismissed):

FelixFan1992 (approved):

PR #43, opened 2025-09-16
Fix: CR FetchPointer able to retrieve CCIPObjectRef

stackman27:


relayer/chainwriter/ptb/offramp/helpers.go
```diff
@@ -327,3 +305,37 @@ func DecodeParameters(lggr logger.Logger, function map[string]any, key string) (
 
 	return paramTypes, nil
 }
+
+func GetOffRampAddressMappingsHelper(ctx context.Context, lggr logger.Logger, ptbClient client.SuiPTBClient, offRampPackageID string, signerAddress string) (string, error) {
```
faisal-chainlink, line 310: It would be super useful to replace with bindings. This way we can remove the custom decoding that's happening below

relayer/chainreader/reader/chainreader.go
```diff
@@ -593,22 +598,50 @@ func (s *suiChainReader) prepareArguments(ctx context.Context, argMap map[string
...
+		}
 		for _, pointer := range pointers {
-			// object tag matches
-			if ownedObject.Data.Type != "" && strings.Contains(ownedObject.Data.Type, pointer) {
-				pointersValuesMap[pointer] = ownedObject.Data.Content.Fields
+			if strings.Contains(obj.Data.Type, pointer) {
+				fields := obj.Data.Content.Fields
+				pointersValuesMap[pointer] = fields
+
+				// cache CCIPObjectRefID
```
faisal-chainlink: Nitpick: we can take this if statement outside of the for loop entirely

The reason being is that we will likely add a few of those checks for each value we are caching

faisal-chainlink (commented):
A couple of minor comments, looks good otherwise

relayer/chainreader/reader/chainreader.go
stackman27: are we planning on caching other values too?

relayer/chainwriter/ptb/offramp/helpers.go
stackman27: done https://github.com/smartcontractkit/chainlink-sui/pull/43/commits/d0094e697110a5b9d1f823aebfd911b2ce7eb8a2

relayer/chainreader/reader/chainreader.go
stackman27: done https://github.com/smartcontractkit/chainlink-sui/pull/43/commits/d0094e697110a5b9d1f823aebfd911b2ce7eb8a2

faisal-chainlink: yes like the state object IDs can be cached since they don't change

faisal-chainlink (approved):

FelixFan1992 (approved):

PR #62, opened 2025-09-24
Fixes from e2e test

stackman27:


contracts/test/Move.lock
```diff
@@ -21,7 +21,7 @@ dependencies = [
 ]
 
 [move.toolchain-version]
-compiler-version = "1.55.0"
+compiler-version = "1.50.1"
```
FelixFan1992: can we revert this?
i think you just need to update your SUI CLI version locally. sth like `brew upgrade sui`. and rebuild/republish this package.

relayer/txm/gas_manager.go
```diff
@@ -138,12 +138,24 @@ func (s *SuiGasManager) EstimateGasBudget(ctx context.Context, tx *SuiTx) (uint6
...
-		if gl, ok := val.(*big.Int); ok {
-			gasLimit.Add(gasLimit, gl)
-		} else {
-			return nil, fmt.Errorf("gasLimit in ExtraArgsDecoded is not *big.Int, got %T", val)
+		var gl *big.Int
+
+		switch v := val.(type) {
+		case *big.Int:
+			gl = v
+		case uint64:
```
FelixFan1992, line 148: when will the gas limit be interpreted as big int and uint64 respectively?

deployment/changesets/cs_mint_token.go
```diff
@@ -0,0 +1,71 @@
...
+		GetCallOpts: func() *bind.CallOpts {
+			b := uint64(400_000_000)
+			return &bind.CallOpts{
+				WaitForExecution: true,
+				GasBudget:        &b,
+			}
+		},
+	}
+
+	// Run MintSuiToken Operation
```
FelixFan1992: mint link token operation

deployment/changesets/cs_mint_token.go
```diff
@@ -0,0 +1,71 @@
...
+	mintLinkTokenReport, err := operations.ExecuteOperation(e.OperationsBundle, linkops.MintLinkOp, deps,
+		linkops.MintLinkTokenInput{
+			LinkTokenPackageId: config.TokenPackageId,
+			TreasuryCapId:      config.TreasuryCapId,
+			Amount:             config.Amount, // 1099999999999999984
+		})
+	if err != nil {
+		return cldf.ChangesetOutput{}, fmt.Errorf("failed to Mint SuiToken for Sui chain %d: %w", config.ChainSelector, err)
```
FelixFan1992: same

deployment/changesets/cs_mint_token.go
FelixFan1992, line 1: technically, this link token will be deployed as the official LINK package on Sui. and we won't mint anything directly on Sui. that said, i understand this is purely for testing purpose.
so it's probably better to eventually switch to MockLink token for any testing.

deployment/changesets/cs_update_prices.go
```diff
@@ -0,0 +1,77 @@
...
+			}
+		},
+	}
+
+	// Run UpdateTokenPrices Operation
+	updateTokenPrices, err := operations.ExecuteOperation(e.OperationsBundle, ccipops.FeeQuoterUpdateTokenPricesOp, deps,
+		ccipops.FeeQuoterUpdateTokenPricesInput{
+			CCIPPackageId: config.CCIPPackageId,
+			CCIPObjectRef: config.CCIPObjectRef,
+			// FeeQuoterCapId:        feeQuoterCapId,
```
FelixFan1992: i don't think we will ever update prices using fee quoter cap with changeset. bc this will only be called by offramp via commit function. 

if this is needed for testing, it can use the function which updates price with owner cap.

relayer/txm/gas_manager.go
stackman27: from core node it's uint64

contracts/test/Move.lock
stackman27: hmm i think it auto added when i ran this test, will revert

deployment/changesets/cs_update_prices.go
stackman27: sounds good, will just remove this entirely. Not really used during testing either

FelixFan1992 (approved):

PR #65, opened 2025-09-25
Chore: re-use the generic balance monitor defined in Aptos

faisal-chainlink:


relayer/monitor/balance.go
```diff
@@ -53,8 +54,12 @@ type balanceClient struct {
 	client client.SuiPTBClient
 }
 
-// GetAccountBalance returns the account balance in APT.
-func (c balanceClient) GetAccountBalance(ctx context.Context, address string) (float64, error) {
+// GetAccountBalance returns the account balance in SUI
+func (c balanceClient) GetAccountBalance(address string) (float64, error) {
+	// TODO: is this safe?
```
faisal-chainlink, line 59: @cfal we need the context for the client call, we can otherwise just use a simple timeout here or alternatively remove this call from the rate limited calls in the client

FelixFan1992 (approved):

PR #70, opened 2025-09-25
feat(decoder): Improve deserializer

rodrigombsoares:
# Summary
The previous way in which the deserializer was structured, we were unable to handle vectors of custom types such as u128, u256. And addresses were being decoded as [32]byte, which is not readable enough for proposal decoding.
This PR re-designed the deserializer logic, while keeping a recursion for vectors and enabling custom decoding for types that are either not supported by sui-sdk or require further transformation.
This PR also renames and exports the global `functioInfo` map on bindgen.

# Features
- Re-design deserializer
- Enable vector<vector<...u128|u256>> decoding
- Decode addresses as `models.SuiAddress` (string)


bindings/bind/bcs_decoder.go
```diff
@@ -299,70 +301,113 @@ func DeserializeBCS(data []byte, moveTypes []string) ([]any, error) {
...
+		return res, typ, err
+	case moveType == "u64":
+		var res uint64
+		typ, err := decode(deserializer, &res)
+		return res, typ, err
+	case moveType == "0x1::string::String":
+		var res string
+		typ, err := decode(deserializer, &res)
+		return res, typ, err
+	case strings.HasPrefix(moveType, "vector"):
```
cfal: nit: check for `HasPrefix("vector<") && HasSuffix(">")`, this prevents breakages eg if another native type with vector prefix is added, we assume this anyway with `moveType[7 : len(moveType)-1]` below

bindings/bind/bcs_decoder.go
```diff
@@ -290,7 +292,7 @@ func DeserializeBCS(data []byte, moveTypes []string) ([]any, error) {
 	deserializer := mystenbcs.NewDecoder(bytes.NewReader(data))
```
cfal: can we add an additional check here that bytes.NewReader.Len() is zero after all the decoding to make sure that all bytes have been consumed as expected?

faisal-chainlink (approved):

PR #74, opened 2025-09-29
Fix: improve relayer withRateLimit semaphore lock acquisition

faisal-chainlink:


relayer/client/ptb_client.go
```diff
@@ -126,18 +127,52 @@ func NewPTBClient(
...
+		if atomic.CompareAndSwapInt32(&released, 0, 1) {
+			c.rateLimiter.Release(1)
+		}
+	}
+
+	// if we haven't returned by the deadline (+ small grace),
+	// reclaim the permit to prevent global deadlock
+	grace := 2 * time.Second
+	_ = time.AfterFunc(c.transactionTimeout+grace, func() {
+		c.log.Warnw("rate limit lease expired; forcing release",
```
cfal: is this necessary? what's a scenario where releaseOnce() wouldn't execute in the defer below?

faisal-chainlink: It's a safe-guard against the function being called not respecting the `timeoutCtx` and possibly holding the lock for too long or even being stuck in a deadlock of some kind.

cfal: how would that occur when there is a `defer releaseOnce()` below?

cfal: this seems to mean that there is no actual rate limit at all, if a function waits for too long, we'll just allow an unlimited number of calls at once. this shouldn't be a black box where we are guessing about the behavior.

faisal-chainlink: I've removed the timer check

stackman27:
can we hold off on merging this until it's e2e tested on staging? will likely lyk by tmrw if this works as expected

FelixFan1992 (approved):

PR #78, opened 2025-10-01
Use `derived_object` in CCIP Router

JohnChangUK:


contracts/ccip/ccip_router/sources/ownable.move
```diff
@@ -56,11 +57,15 @@ const ETransferNotAccepted: u64 = 8;
 const ECannotTransferToMcms: u64 = 9;
 const EMustTransferToMcms: u64 = 10;
 
-public fun new(ctx: &mut TxContext): (OwnableState, OwnerCap) {
+public fun new<K: copy + drop + store>(
+    uid: &mut UID,
+    key: K,
```
cfal, line 73: as discussed, perhaps this can just be a string.

also, maybe we can have:
`new(uid, ctx)` - which defaults the key to something like "CCIP_OWNABLE"
and
`new_with_custom_key(uid, "RouterKeyOwnerCap", ctx)`

wdyt?

contracts/ccip/ccip_router/sources/router.move
```diff
@@ -40,31 +48,35 @@ const EInvalidObjectAddress: u64 = 6;
 const EInvalidOnrampAddress: u64 = 7;
 
 fun init(_witness: ROUTER, ctx: &mut TxContext) {
-    let (ownable_state, owner_cap) = ownable::new(ctx);
+    let mut router_object = RouterObject { id: object::new(ctx) };
+    let (ownable_state, owner_cap) = ownable::new(&mut router_object.id, RouterOwnableKey(), ctx);
 
     let router = RouterState {
-        id: object::new(ctx),
+        id: derived_object::claim(&mut router_object.id, RouterKey()),
```
cfal: ditto on string if it works

relayer/codec/derived_object.go
```diff
@@ -0,0 +1,119 @@
+package codec
```
cfal: these files should be in bindings/bind - we need to use these in tests/changesets that do not have relayer as a dependency

contracts/ccip/ccip_router/sources/ownable.move
JohnChangUK: Changed to `vector<u8>`.  And added the default

contracts/ccip/ccip_router/sources/router.move
JohnChangUK: Amended to `vector<u8>`

relayer/codec/derived_object.go
JohnChangUK: Good point, moved.

faisal-chainlink (approved):

PR #79, opened 2025-10-01
Add ptb client documentation

JoaoVasques:
Generated during the knowledge transfer session.


faisal-chainlink:
Thanks for adding this @JoaoVasques I need to make a small update to the client and I'll include it in the docs here before merging the PR

faisal-chainlink (approved):

PR #80, opened 2025-10-01
Add MCMS dynamic proposal generator seq

RodrigoAD:
- [x] Test sample proposal e2e

Next PR
- [ ] Convert every operation needed for MCMS to export `Call`
- [ ] Export every operation needed for MCMS in a single batch that can be registered in `OperationRegistry`
- [ ] Documentation generator for Operation definitions and Input samples that are needed for MCMS


deployment/changesets/cs_mcms_proposal_generate.go
```diff
@@ -0,0 +1,46 @@
...
+
+func (d MCMSProposalGenerate) Apply(e cldf.Environment, config mcmsops.ProposalGenerateInput) (cldf.ChangesetOutput, error) {
+	suiChains := e.BlockChains.SuiChains()
+
+	suiChain := suiChains[config.ChainSelector]
+	deps := sui_ops.OpTxDeps{
+		Client: suiChain.Client,
+		Signer: nil, // Signer is not needed since we are not executing any transactions
+		GetCallOpts: func() *bind.CallOpts {
+			b := uint64(500_000_000)
```
rodrigombsoares: Is it always this value? Or should this be an input?

deployment/ops/mcms/op_proposal_generate.go
```diff
@@ -0,0 +1,142 @@
...
+
+	// Proposal related
+	Role  suisdk.TimelockRole `json:"role"`
+	Delay time.Duration       `json:"delay"`
+
+	// Chain related
+	ChainSelector uint64 `json:"chainSelector"`
+}
+
+var GenerateProposalHandler = func(b cld_ops.Bundle, deps sui_ops.OpTxDeps, input ProposalGenerateInput) (output mcms.TimelockProposal, err error) {
```
rodrigombsoares: nit: why is it a var instead of func `generateProposalHandler`?

rodrigombsoares: non nit: no need to export

deployment/ops/mcms/op_proposal_generate.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+	return *timelockProposal, nil
+}
+
+var MCMSDynamicProposalGenerateSeq = cld_ops.NewSequence(
+	sui_ops.NewSuiOperationName("mcms", "proposal", "generate"),
+	semver.MustParse("0.1.0"),
+	"Generates an MCMS timelock proposal that batches multiple operations based on the provided definitions and inputs",
+	GenerateProposalHandler,
```
rodrigombsoares: Usually in go I see functions declarations in call order, so I'd do:
```
var MCMSDynamicProposalGenerateSeq ...
...
func generateProposalHandler(...) {...}
```
ref: https://github.com/uber-go/guide/blob/master/style.md#function-grouping-and-ordering

deployment/ops/mcms/op_proposal_generate.go
```diff
@@ -0,0 +1,142 @@
...
+		)
+		mcmsTxs[i] = tx
+	}
+
+	op := types.BatchOperation{
+		ChainSelector: types.ChainSelector(input.ChainSelector),
+		Transactions:  mcmsTxs,
+	}
+
+	validUntilMs := uint32(time.Now().Add(time.Hour * 24).Unix())
```
rodrigombsoares: Default is 72h
https://github.com/smartcontractkit/chainlink/blob/099c08ad7f80d48fcf38552d24f9eaa5435c3e1f/deployment/common/proposalutils/propose.go#L29

deployment/ops/mcms/op_proposal_generate.go
```diff
@@ -0,0 +1,142 @@
...
+	metadata, err := suisdk.NewChainMetadata(0, input.Role, input.MmcsPackageID, input.McmsStateObjID, input.AccountObjID, input.RegistryObjID, input.TimelockObjID)
+	if err != nil {
+		return mcms.TimelockProposal{}, fmt.Errorf("failed to create chain metadata: %w", err)
+	}
+
+	var action types.TimelockAction
+	var delay *types.Duration
+	switch input.Role {
```
rodrigombsoares, line 98: Isn't there a `timelockCanceller` role?

deployment/ops/ccip/op_state_object.go
rodrigombsoares: Do we need to call another `AcceptOwnership` method on bindings? Can't we just execute the `call` that was previously generated?
This method does the same call to `Encoder().AcceptOwnership` and then executes transaction. 
I guess we could just `contract.ExecuteTransaction(ctx, opts, encodedCall)` and avoid encoding twice.
This would also allow for a helper method to "execute or return call" that could be used on every Op

rodrigombsoares: Encode in the package, call the helper method and return. With some generics

deployment/changesets/cs_mcms_proposal_generate.go
RodrigoAD: It's actually not used, this isn't sending any tx, I will remove it

deployment/ops/mcms/op_proposal_generate.go
RodrigoAD: > nit: why is it a var instead of func generateProposalHandler?

To keep consistency with the rest of operations, could also be a `func`

deployment/ops/mcms/op_proposal_generate.go
RodrigoAD: Yes there is! Thanks, totally missed it

deployment/ops/ccip/op_state_object.go
RodrigoAD: Totally right, I updated it to use the contract method `Bound().ExecuteTransaction` that accepts the `encodedCall`, builds the PTB and executes

deployment/ops/mcms/op_proposal_generate.go
RodrigoAD: This follows the rest of operation standards, leaving as is. We can reconsider this later for every Operation

rodrigombsoares (approved):

JohnChangUK (approved):

PR #81, opened 2025-10-01
Chore: separate client instances for indexers

faisal-chainlink:
Depends on #74 

---

This PR instantiates the PTBClient separately for the use of indexers to avoid congesting the same rate limiter.


JohnChangUK (dismissed):

relayer/plugin/relayer.go
```diff
@@ -95,6 +95,22 @@ func NewRelayer(cfg *config.TOMLConfig, lggr logger.Logger, keystore core.Keysto
...
+		// frequent RPC calls to get latest values
+		maxConcurrentRequests*3,
+		client.TransactionRequestType(requestType),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("error in NewRelayer (monitor): %w", err)
+	}
+
+	// Use a separate client for the indexers to avoid rate limiting
+	suiClientIndexers, err := client.NewPTBClient(
```
stackman27, line 113: @faisal-chainlink  could you pls expand on this change? How does using seperate client here avoid rateLimit?

stackman27, line 113: @faisal-chainlink i wonder if this causes issues when as  this function returns a client as well in `SuiRelayer {} struct` how do you decide which one to return?

faisal-chainlink: Writing a comment here based on our in-call discussion: 

Primarily it's the indexers that use a different client to avoid using the same allocation from the semaphore, even if it doesn't cause the issue, it can still be considered a nice to have

FelixFan1992 (approved):

PR #84, opened 2025-10-02
Use derived_object in CCIP, Onramp, Offramp

JohnChangUK:
## Desc
Derived objects for CCIP, Onramp, Offramp

### On chain
Object addresses are derived deterministically. 
- `Object ID` + `vector<u8>` key

### Off chain
- `PointerTag` now a struct, added PackageID field to PointerTag
- Allows specifying cross-package dependencies explicitly
- Used when pointer module differs from calling contract's package

```
pointerTag := &codec.PointerTag{
    Module:        "state_object",
    PointerName:   "CCIPObjectRefPointer", 
    FieldName:     "ccip_object_id",
    DerivationKey: "CCIPObjectRef",
    PackageID:     secondaryPackageId,  // Cross-package reference
}
```


relayer/chainreader/reader/chainreader.go
```diff
@@ -558,8 +560,8 @@ func (s *suiChainReader) prepareArguments(ctx context.Context, argMap map[string
 		return []any{}, []string{}, nil
 	}
 
-	// referring to the tag parts "_::module::Pointer::field"
-	tagLength := 4
+	// referring to the tag parts "_::module::Pointer::field::derivationKey"
```
faisal-chainlink: What do you think about changing the tag to be a struct instead? 

Something like the following:

```
somePointerTag := PointerTag {
    tag: "_::module::Pointer::field",
    derivationPath: "..."
}
```

relayer/chainreader/reader/chainreader.go
```diff
@@ -605,35 +607,41 @@ func (s *suiChainReader) prepareArguments(ctx context.Context, argMap map[string
...
+		pointerFieldValues := make(map[string]string)
+		for _, pointerVal := range pointerVals {
+			derivedId, err := bind.DeriveObjectIDWithVectorU8Key(parentObjectId, []byte(pointerVal.derivationKey))
+			if err != nil {
+				return nil, nil, fmt.Errorf("failed to derive object ID for %s using key %s: %w", pointerVal.field, pointerVal.derivationKey, err)
+			}
+			pointerFieldValues[pointerVal.paramName] = derivedId
 		}
 
 		// add the values to the arg map
```
faisal-chainlink: Can this loop be merged into the one above? They both iterate over the same `pointerVals`

faisal-chainlink (commented):
I'm wondering if it's better for us to simply load all the object IDs from all the pointers ahead of time rather than get them from the RPC each time. 

This would actually mean that we would reduce the number of RPC calls and potentially reduce the code complexity. 

Right now we are coupling the flow of getting the parent object ID with the process of derivation but those could actually be done separately.

Given that we already have a map in `GetParentObjectID`

```
parentFieldMap := map[string]string{
	"OffRampStatePointer":  "off_ramp_object_id",
	"OnRampStatePointer":   "on_ramp_object_id",
	"CCIPObjectRefPointer": "ccip_object_id",
	"RouterStatePointer":   "router_object_id",
	"CounterPointer":       "counter_object_id", // Test contract: parent object ID
}
```

As soon as the binding takes place with the `OffRamp` package ID, we should be able to load all of them, right?

relayer/chainreader/reader/chainreader.go
JohnChangUK: Yup

PR #89, opened 2025-10-03
Add wallet creation script

RodrigoAD:
Adds a script and task to create a fresh wallet and exported as a Hex address


faisal-chainlink (approved):

faisal-chainlink:
Do we need to also set this wallet as the `active-address` after it's generated in the environment?

RodrigoAD:
> Do we need to also set this wallet as the `active-address` after it's generated in the environment?

The purpose is just to generate and export a new wallet. Users can always activate it as default after is generated

PR #92, opened 2025-10-03
Fix: add check for 0x0 receiver addresses in offramp flow

faisal-chainlink:


relayer/chainwriter/ptb/offramp/execute.go
```diff
@@ -352,6 +352,13 @@ func ProcessReceivers(
 			continue
 		}
 
+		// Check if receiver is a zero address (0x0....0 // 32 bytes of 0)
```
cfal, line 357: the above log message for message.Receiver == nil should be at least lggr.Errorw with a better error message, as it is unexpected as we know the receiver is always deserialized into 32 bytes

relayer/chainwriter/ptb/offramp/execute.go
```diff
@@ -352,6 +352,13 @@ func ProcessReceivers(
...
+		receiverHex := hex.EncodeToString(message.Receiver)
+		if len(receiverHex) == 64 && receiverHex == strings.Repeat("0", 64) {
```
cfal: can we do a slice/byte comparison? encoding to a string only for comparison seems strange. eg you can do `bytes.Equal(message.Receiver, types.AccountZero)` - see AccountZero in the Aptos SDK, or if an equivalent exists in the Sui SDK (or define it ourselves)

faisal-chainlink: This was updated in the same moment as the comment I think 😅

faisal-chainlink: Will add the `AccountZero` type though

relayer/common/utils.go
```diff
@@ -85,3 +86,7 @@ func ConvertBytesToHex(value any) any {
 		return value
 	}
 }
+
+func IsZeroAddress(address []byte) bool {
+	return len(address) == 32 && bytes.Equal(address, make([]byte, 32))
```
cfal: could we make it a const instead of allocating a new array each comparison?

relayer/chainwriter/ptb/offramp/execute.go
cfal: or you can just reference types.AccountZero since aptos-go-sdk is imported

relayer/chainwriter/ptb/offramp/execute.go
faisal-chainlink: Changed to Errorw log and changed the text

JohnChangUK (approved):

relayer/common/utils.go
faisal-chainlink: already added ✅

PR #96, opened 2025-10-06
update hot potato creation

FelixFan1992:


contracts/ccip/ccip/sources/onramp_state_helper.move
```diff
@@ -41,7 +41,7 @@ fun init(_witness: ONRAMP_STATE_HELPER, ctx: &mut TxContext) {
     transfer::transfer(source_cap, ctx.sender());
 }
 
-public fun create_token_transfer_params(token_receiver: address): TokenTransferParams {
+public fun create_token_transfer_params(token_receiver: vector<u8>): TokenTransferParams {
```
faisal-chainlink, line 45: Can we validate the address since it's a method that's callable publicly?

FelixFan1992: added a check for length. no additional checks are possible at this moment bc we don't know the destination chain and if this message has a token transfer right now.

JohnChangUK (approved):

PR #99, opened 2025-10-08
Added GetFeeComponents

stackman27:


relayer/chainwriter/chainwriter.go
```diff
@@ -136,7 +136,16 @@ func (s *SuiChainWriter) SubmitTransaction(ctx context.Context, contractName str
 
 // GetFeeComponents implements types.ContractWriter.
 func (s *SuiChainWriter) GetFeeComponents(ctx context.Context) (*commonTypes.ChainFeeComponents, error) {
```
FelixFan1992, line 138: is there a place where we can read what these endpoints are supposed to do? i don't have the context of what fee component is

faisal-chainlink: This is called by the core node directly to create more detailed estimates

faisal-chainlink (approved):

PR #100, opened 2025-10-08
get token receiver and gas limit for Sui destination chain

FelixFan1992:


JohnChangUK (approved):

contracts/ccip/ccip/sources/fee_quoter.move
```diff
@@ -845,6 +846,21 @@ fun resolve_generic_gas_limit(dest_chain_config: &DestChainConfig, extra_args: v
...
 }
 
+fun resolve_sui_gas_limit(dest_chain_config: &DestChainConfig, extra_args: vector<u8>): u256 {
+    let (
+        gas_limit,
+        allow_out_of_order_execution,
+        _token_receiver,
+        _receiver_object_ids,
+    ) = decode_sui_extra_args(extra_args);
+    assert!(gas_limit <= (dest_chain_config.max_per_msg_gas_limit as u64), EMessageGasLimitTooHigh);
```
JohnChangUK, line 856: Shouldn't downcast `dest_chain_config.max_per_msg_gas_limit` as it may lose precision. We can cast up `gas_limit` instead

FelixFan1992: why this is downcast?
gas_limit is u64 and max_per_msg_gas_limit is u32. so by casting it to u64, it should not lose precision?

JohnChangUK: Ah you're right, ignore

PR #106, opened 2025-10-09
simplify the set_pool in token admin registry

FelixFan1992:


contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -445,19 +445,19 @@ fun unregister_pool_internal(
...
-    token_pool_module: String,
     lock_or_burn_params: vector<address>,
     release_or_mint_params: vector<address>,
     _: TypeProof,
     ctx: &mut TxContext,
 ) {
-    let token_pool_type_proof_tn = type_name::with_defining_ids<TypeProof>();
-    let token_pool_type_proof_str = type_name::into_string(token_pool_type_proof_tn);
+    let proof_tn = type_name::with_defining_ids<TypeProof>();
+    let proof_package_id = address::from_ascii_bytes(&proof_tn.address_string().into_bytes());
```
JohnChangUK, line 454: We can always derive the token pool package ID from the proof, is there a reason to store them both?

FelixFan1992: mostly for convenience

JohnChangUK (approved):

PR #108, opened 2025-10-09
fix(bindings): registryObjID as NewCCIPEntrypointArgEncoder args

rodrigombsoares:
# Summary
Fixes `NewCCIPEntrypointArgEncoder` to accept a `registryObjID` as parameter. Otherwise, since `registryObjID` is unexported, it'd be impossible to initialize `&CCIPEntrypointArgEncoder{registryObjID: registryObjID}`.


RodrigoAD (approved):
Thanks @rodrigombsoares !

PR #109, opened 2025-10-10
move token pool logic into concrete token pool implementations

FelixFan1992:


contracts/ccip/ccip_token_pools/burn_mint_token_pool/sources/ownable.move
```diff
@@ -56,7 +54,7 @@ const ETransferNotAccepted: u64 = 8;
 const ECannotTransferToMcms: u64 = 9;
 const EMustTransferToMcms: u64 = 10;
 
-public fun new(ctx: &mut TxContext): (OwnableState, OwnerCap) {
+public(package) fun new(ctx: &mut TxContext): (OwnableState, OwnerCap) {
```
JohnChangUK, line 57: Thanks for changing `ownable` module too. Let's make all the public functions `package` protected too.

faisal-chainlink (approved):
Thanks for reflecting the changes in the relayer as well. Can we also update the script to deploy the onramp methods for the docs team?

PR #110, opened 2025-10-10
Chore: misc tests cleanup

faisal-chainlink:
* Remove stale methods in tests for env setup
* Update bindings to use a timeout client for the SDK
* Misc cleanup


JohnChangUK (dismissed):

integration-tests/offramp/execute_test.go
```diff
@@ -64,7 +69,9 @@ func setupClients(t *testing.T, lggr logger.Logger) (rel.SuiSigner, sui.ISuiAPI,
 	})
 
 	// Create the client.
-	client := sui.NewSuiClient(testutils.LocalUrl)
+	client := sui.NewSuiClientWithCustomClient(testutils.LocalUrl, &http.Client{
```
RodrigoAD: Why do we have different `setupClients` for offramp and onramp? And why doesn't this use the `bindings/testenv` client setup?

bindings/tests/testenv/testenv.go
```diff
@@ -192,7 +194,9 @@ func (te *TestEnvironment) cleanup() {
 }
 
 func createClient() sui.ISuiAPI {
-	return sui.NewSuiClient(fmt.Sprintf("http://localhost:%d", instance.rpcPort))
+	return sui.NewSuiClientWithCustomClient(fmt.Sprintf("http://localhost:%d", instance.rpcPort), &http.Client{
+		Timeout: HTTPClientTimeout,
+	})
```
RodrigoAD, line 203: You could add an optional timeout param
```suggestion
func createClient(timeout ...time.Duration) sui.ISuiAPI {
    clientTimeout := HTTPClientTimeout
    if len(timeout) > 0 {
        clientTimeout = timeout[0]
    }
    
    return sui.NewSuiClientWithCustomClient(
        fmt.Sprintf("http://localhost:%d", instance.rpcPort), 
        &http.Client{Timeout: clientTimeout},
    )
```

integration-tests/offramp/execute_test.go
faisal-chainlink: This whole test is outdated and needs to be updated. It's currently being ignored but I agree they should both go through the same setup. There's a unified setup used in the relayer already (from testutils), no need to use the one from bindings.

bindings/tests/testenv/testenv.go
faisal-chainlink: Applied

integration-tests/offramp/execute_test.go
faisal-chainlink: Removed this test in favor of the E2E test now running as part of CI

PR #115, opened 2025-10-11
e2e core+relayer test in ci

stackman27:
core ref: 056dd5c326a68277522a8c9cca7be5dd5ebf07ef


stackman27:
Blocked by https://github.com/smartcontractkit/chainlink/pull/19848

cfal (dismissed):
would be great to follow up with saving the artifacts so that the errors can be debugged as well, think we discussed briefly in the internal PR.

stackman27:
> would be great to follow up with saving the artifacts so that the errors can be debugged as well, think we discussed briefly in the internal PR.

ah right! added

faisal-chainlink (dismissed):

faisal-chainlink (dismissed):

FelixFan1992 (approved):

PR #119, opened 2025-10-14
Add type proof to MCMS ExecutingCallbackParams

JohnChangUK:
## Desc
Add type proof to MCMS ExecutingCallbackParams


faisal-chainlink:
@JohnChangUK heads up that the current E2E CI workflow times out at 10 minutes, can we also change this to be a bit longer since E2E tests can be a bit slow when running in parallel. Perhaps 15 minutes is sufficient.

One of the E2E tests were failing and re-running that failed CI task alone caused it to complete successfully in under 10 minutes.

FelixFan1992 (dismissed):

faisal-chainlink (approved):

PR #127, opened 2025-10-15
Fix MCMS accept ownership data decoding

RodrigoAD:


JohnChangUK (approved):

faisal-chainlink:
@RodrigoAD heads up that I merged `develop` into this to check if the extended time for the tests is resolved. There's seems to be another issue causing the unit tests of the contracts to fail.

JohnChangUK (changes requested):
Check if we can generate with `ownable` bindings

RodrigoAD:
Binding approach works

PR #130, opened 2025-10-16
Fix mcms user contracts and tests

RodrigoAD:
### Why
`mcms_user` contracts were not compiling. CI didn't catch it as they are not tested

### What
- Fix the `mcms_user` contracts
- Adds them to the contracts test script


contracts/scripts/test.sh
```diff
@@ -9,6 +9,8 @@ sui move test --path ccip/ccip_token_pools/managed_token_pool
 sui move test --path ccip/ccip_token_pools/lock_release_token_pool
 sui move test --path ccip/ccip_token_pools/burn_mint_token_pool
 sui move test --path mcms/mcms
+sui move test --path mcms/mcms_test
```
RodrigoAD, line 12: These contracts were missing in the CI testing

JohnChangUK (approved):

PR #161, opened 2025-10-23
store and check token pool package id to coin metadata

FelixFan1992:


JohnChangUK (approved):

contracts/ccip/ccip/sources/token_admin_registry.move
```diff
@@ -398,6 +401,13 @@ fun register_pool_internal(
     };
 
     state.token_configs.push_back(coin_metadata_address, token_config);
+    assert!(
+        !state.token_pool_package_id_to_coin_metadata.contains(token_pool_package_id),
```
JohnChangUK, line 405: Perhaps we should consider object state ID instead.

PR #171, opened 2025-10-29
Transfer UpgradeCap with OwnerCap, Use LinkedTable for signed hashes

JohnChangUK:
## MCMS
- Transfer `UpgradeCap` along with `OwnerCap` when transferring out of MCMS (when `mcms_execute_ownership_transfer` is called)
- Use `LinkedTable` for signed hashes storage in `mcms.move`

In  `mcms_execute_ownership_transfer`, the current/latest package address is bcs serialized as there is no way to get the current package address in the Move contract.

## Bindings Update
`mcms_execute_ownership_transfer` takes in `DeployerState`, so MCMS lib needs updating


contracts/ccip/ccip/sources/state_object.move
```diff
@@ -311,6 +321,7 @@ public fun mcms_accept_ownership(
 public fun mcms_execute_ownership_transfer(
     ref: &mut CCIPObjectRef,
     registry: &mut Registry,
+    deployer_state: &mut DeployerState,
```
RodrigoAD, line 324: I think is the first time we'll need the `DeployerState`, we don't currently store it in CLD. 

Do you mind adding it to the address book [here](https://github.com/smartcontractkit/chainlink-sui/blob/develop/deployment/changesets/cs_mcms_deploy.go#L99)? Same as other objects. It's available in the `mcmsReport.Objects`

JohnChangUK: Will do, thanks!

RodrigoAD (dismissed):
Thanks @JohnChangUK 👍

contracts/ccip/ccip/sources/state_object.move
```diff
@@ -334,6 +345,16 @@ public fun mcms_execute_ownership_transfer(
     bcs_stream::assert_is_consumed(&stream);
 
     let owner_cap = mcms_registry::release_cap(registry, McmsCallback {});
+
+    if (mcms_deployer::has_upgrade_cap(deployer_state, @ccip)) {
```
FelixFan1992: did you test this? in ccip package, i am not sure `@ccip` will give you the package id.
another thing is if a package is upgraded, should we still use the old package id as the key in deployer state? if so, we probably need to get it in a different way.

JohnChangUK: We have [state_object_upgrade_test](https://github.com/smartcontractkit/chainlink-sui/pull/171/files#diff-4407950ea6cbf308e85ec5986132c2a93930a03f2acd2a78062c7589ba9ca1d7) which tests this.

If the package is upgraded, the [new package address is registered.](https://github.com/smartcontractkit/chainlink-sui/blob/develop/contracts/mcms/mcms/sources/mcms_deployer.move#L120)

JohnChangUK: We now check the current package address. This is deserialized from MCMS bcs stream

FelixFan1992 (approved):

PR #173, opened 2025-10-29
fix usdc token pool

FelixFan1992:


contracts/ccip/ccip_token_pools/usdc_token_pool/sources/usdc_token_pool.move
```diff
@@ -1160,7 +1160,7 @@ public fun mcms_execute_ownership_transfer<T>(
 ) {
     let (_owner_cap, function, data) = mcms_registry::get_callback_params_with_caps<
         McmsCallback,
```
FelixFan1992, line 1162: @JohnChangUK i realize in other token pools, this mcms call back is typed. should we do that here too?

JohnChangUK: Yes we should

JohnChangUK (approved):

PR #181, opened 2025-10-30
Sui Max Data

stackman27:
core ref: f986aede32a4cf47082ae147c667ec94cf5c9daf


deployment/config.go
```diff
@@ -25,7 +25,7 @@ var DefaultCCIPSeqConfig = ccipops.DeployAndInitCCIPSeqInput{
 	// Fee Quoter destination chain configuration
 	IsEnabled:                         true,
 	MaxNumberOfTokensPerMsg:           1,
-	MaxDataBytes:                      30_000,
+	MaxDataBytes:                      16_000, // the maximum size of a single pure argument in Sui is 16KB
```
JohnChangUK, line 28: Isn't this max bytes per transaction?

stackman27: for pure arg size 

https://move-book.com/guides/building-against-limits/#single-pure-argument-size

JohnChangUK (approved):

PR #187, opened 2025-10-31
Normalize ms timestamp in Sui DB queries 

stackman27:


relayer/chainreader/database/utils.go
```diff
@@ -37,10 +37,19 @@ func BuildSQLCondition(expr query.Expression, args *[]any, argCount *int) (strin
...
 		case *primitives.Timestamp:
+
+			// Detect and normalize millisecond timestamps if there is any.
+			// Sui use ms since epoch, while DB expects seconds.
+			// Using 9_999_999_999 (~year 2286) as a cutoff: anything larger is clearly ms.
+			ts := v.Timestamp
+
+			if ts > 9999999999 {
+				ts = ts / 1000
+			}
```
faisal-chainlink, line 48: What's the value in adding this `if` statement? Can't we basically work with the value as if always being divisible by 1000?

stackman27: We’ll likely never hit this condition since we now insert timestamps in seconds into the DB. This check was added as a cleanup safeguard for any existing millisecond data and as a general safety net for future inconsistencies.

faisal-chainlink (dismissed):

FelixFan1992 (approved):

PR #192, opened 2025-11-01
Add MCMS and other functions

JohnChangUK:
core ref: 9d93d9a8078e97a6049ce37afe05b30847600a43


contracts/ccip/ccip_onramp/sources/onramp.move
```diff
@@ -1423,67 +1423,6 @@ public fun mcms_execute_ownership_transfer(
     execute_ownership_transfer(ref, owner_cap, state, to, ctx);
 }
 
-public fun mcms_initialize(
```
JohnChangUK, line 1426: Not needed as MCMS doesn't store the nonce manager and source transfer caps.

FelixFan1992 (approved):

PR #201, opened 2025-11-03
Update event indexer and txn indexer to insert txDigest as a hex string

faisal-chainlink:


relayer/chainreader/indexer/transactions_indexer.go
```diff
@@ -431,13 +432,27 @@ func (tIndexer *TransactionsIndexer) syncTransmitterTransactions(ctx context.Con
 				continue
 			}
 
+			// Convert the txDigest to hex
+			txDigestHex := transactionRecord.Digest
+			if base64Bytes, err := base64.StdEncoding.DecodeString(txDigestHex); err == nil {
+				hexTxId := hex.EncodeToString(base64Bytes)
+				txDigestHex = "0x" + hexTxId
+			}
```
FelixFan1992, line 449: should we consider what happens when err is not nil?

FelixFan1992: or we can safely assume that `transactionRecord.Digest` is going to be valid

faisal-chainlink: I think we can safely assume it is always there. The thing that I want to check is if the core node currently assumes this to be a base64 string when getting events, if that's the case, we need to make a small change in core as well. cc @stackman27

FelixFan1992: ok let's wait for his input

stackman27: Since both fields are expected as bytes, i think decoded hex is better 

```
type Head struct {
	Height string
	Hash   []byte
	// Timestamp is in Unix time
	Timestamp uint64
}

type Sequence struct {
	// This way we can retrieve past/future sequences (EVM log events) very granularly, but still hide the chain detail.
	Cursor string
	TxHash []byte
	Head
	Data any
}
```

faisal-chainlink: @pedrovsg needs it as a string here, let's make a decision and either merge or close this PR @stackman27

stackman27 (approved):
LGTM! hex is better than base64

faisal-chainlink:
Moved to #257

PR #203, opened 2025-11-03
apply several small fixes

FelixFan1992:


contracts/mcms/mcms/sources/bcs_stream.move
FelixFan1992, line 1: format this file to Sui convention and move tests to its own test file

JohnChangUK (approved):

PR #212, opened 2025-11-05
Fix ChainReader GLV for upgrades

stackman27:
core ref: 701e659e63b1b969bd4ebc4bb3eb6489577885dd

### DIFF
--- a/relayer/chainreader/reader/chainreader.go
+++ b/relayer/chainreader/reader/chainreader.go
@@ -14,7 +14,6 @@ import (
 
 	"github.com/smartcontractkit/chainlink-common/pkg/services"
 	"github.com/smartcontractkit/chainlink-common/pkg/sqlutil"
-	"github.com/smartcontractkit/chainlink-common/pkg/types"
 
 	"github.com/smartcontractkit/chainlink-aptos/relayer/chainreader/loop"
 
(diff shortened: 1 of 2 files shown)


relayer/chainreader/reader/chainreader.go
```diff
@@ -239,7 +240,44 @@ func (s *suiChainReader) GetLatestValue(ctx context.Context, readIdentifier stri
...
+		newBinding := types.BoundContract{
+			Name:    contractName,
+			Address: parsed.address, // Package ID of the deployed counter contract
+		}
+
+		err := s.Bind(ctx, []pkgtypes.BoundContract{newBinding})
+		if err != nil {
+			return fmt.Errorf("error while binding new contract: %w", err)
+		}
+	}
```
faisal-chainlink: There are already helper methods to achieve this type of thing. I'd recommend we re-use them from the client instead of writing new code here.

relayer/chainreader/reader/chainreader.go
```diff
@@ -239,7 +240,44 @@ func (s *suiChainReader) GetLatestValue(ctx context.Context, readIdentifier stri
...
+
+	// this is the upgraded pkgID
+	parsed.address = latestPackageId
+
+	s.logger.Info("LATEST PKG ID POST GETLATESTPKGID: ", parsed.address)
+
+	// Bind the new pkgID if the latestPkgID is not currentPkgId
+	// this means there has been an upgrade
+	if originalPkgId != latestPackageId {
```
faisal-chainlink: Why do we need to re-bind the latest package ID?

relayer/chainreader/reader/chainreader.go
```diff
@@ -239,7 +240,44 @@ func (s *suiChainReader) GetLatestValue(ctx context.Context, readIdentifier stri
...
+	s.logger.Info("GLV UPGRADE: ", parsed.address, parsed.contractName)
+
+	module := "state_object"
+	// anything under ccip module has to be state_object module
+	switch parsed.contractName {
+	case "OffRamp":
+		module = "offramp"
+	case "OnRamp":
+		module = "onramp"
+	}
```
faisal-chainlink: If this is necessary, can we put it in a helper in `/relayer/common`

faisal-chainlink (changes requested):

faisal-chainlink:
I'd recommend a simpler approach overall here. Simply calling `GetLatestPackageId` in the `executeFunction` method before actually making the client call is sufficient to replace the package ID with the latest one.

faisal-chainlink (dismissed):

JohnChangUK (approved):

PR #221, opened 2025-11-06
Fix: Improve Sui address validation

faisal-chainlink:
The SDK only verifies if the provided address starts with `0x` and has a total of 32 bytes. We need to also ensure that it's valid hex.

This PR creates a simple helper in bindings to better validate Sui addresses. It is also re-used in the Sui relayer.


bindings/bind/utils.go
```diff
@@ -1,16 +1,55 @@
...
+			return false
+		}
+	}
+
+	// hex.DecodeString requires even length; allow odd by left-padding one '0'
+	if len(h)%2 == 1 {
+		h = "0" + h
+	}
+
+	b, err := hex.DecodeString(h)
```
FelixFan1992, line 39: if an input is `0x6`, it will be transformed to `06` at this point, can it be decoded to 32 bytes?
or should we left pad to 32 bytes at line 36?

faisal-chainlink: nice catch, I think left padding to 32 bytes would be best

faisal-chainlink: On second thought, we don't actually need to do this. That's because this method should still return `true` for the value of `0x6` or `0x06`. It will simply validate that it's less than 32 bytes. 

There's another method in this file (`#ToSuiAddress(..)`) which handles left padding when necessary using the SDK.

FelixFan1992 (approved):

PR #226, opened 2025-11-09
Enable Sui Upgrades 

stackman27:


faisal-chainlink (approved):

contracts/ccip/ccip/Move.toml
```diff
@@ -3,7 +3,7 @@ name = "ChainlinkCCIP"
 edition = "2024"
 
 [addresses]
-ccip = "_"
+ccip = "0x0"
```
FelixFan1992, line 6: not blocking. but these are address changes necessary?

stackman27: yea needed this to compile now 

i've modified the script to run the contract test: https://github.com/smartcontractkit/chainlink-sui/blob/ca925727b5b4f1171d20461241ca728f7c22b475/contracts/scripts/test.sh

PR #227, opened 2025-11-10
Sui views

rodrigombsoares:
# Summary
This PR adds views generator for SUI. This is what is used to generate `state.json` files in cld (eg: [aptos-state](https://github.com/smartcontractkit/chainlink-deployments/blob/main/domains/ccip/aptos-staging/state.json)) consumed by docs and o11y.

# Features
- Added integration tests that use changesets in `integration-tests/deploy/`
- Implemented views in `deployment/view/`
- Implemented chain view generator in `deployment/state.go`

# Suggestion for reviewers
1. Start by reviewing the tests
2. Go over the main view generator in `deployment/state.go`
3. Review each view generator inside `deployment/view/`

# Testing
- integration tests added
- test using contracts on cld
  - create a workspace with `chainlink-deployments`, `chainlink` and `chainlink-sui`
  - run command `state generate -e sui-staging` from `chainlink/deployment/domains/ccip/cmd`
  - `chainlink` should be on branch [wip]


faisal-chainlink (dismissed):
Looks good, just a couple of minor things: 

1. Need to do `go mod tidy` for CI to pass
2. Is there a scenario where we want to get the state object IDs within the method to generate the views? Or will we always generate the views once after the deployment and that's it?
3. Do you think there's a benefit to making the state object IDs available in the views?

deployment/view/offramp.go
```diff
@@ -0,0 +1,136 @@
...
+
+// GenerateOffRampView generates an offramp view for a given offramp by querying the on-chain state
+func GenerateOffRampView(
+	ctx context.Context,
+	chain sui.Chain,
+	offRampPackageID string,
+	offRampStateObjectID string,
+	ccipObjectRef string,
+) (OffRampView, error) {
+	if offRampPackageID == "" || offRampStateObjectID == "" {
```
faisal-chainlink, line 49: Do you see any benefit to having the `offrampStateObjectID` deduced from the contracts as a fallback? I.e. is there a valid scenario where we would not know the state object ID but still want to move forward? 

The same question applies to the `ccipObjectRef` value.

rodrigombsoares: https://smartcontract-it.atlassian.net/browse/NONEVM-3043

FelixFan1992 (approved):

PR #235, opened 2025-11-10
Enhancement: check failed transaction's package ID

faisal-chainlink:
This PR checks the package ID in a failed transaction to ensure that a malicious transmitter does not a failure incorrectly.


faisal-chainlink:
Note: The assumption is that we only need to check against a single (latest) package ID here as the offramp execution runs against the latest package ID (always).

FelixFan1992 (approved):

PR #238, opened 2025-11-11
Fix: Update TXM broadcaster to check the status field of broadcast attempts

faisal-chainlink:
This PR adds a change to check that the `Status` field of the response is empty which will be the case where errors are not related to an actual transaction failure.

### DIFF
--- a/relayer/txm/broadcaster.go
+++ b/relayer/txm/broadcaster.go
@@ -89,13 +89,10 @@ func broadcastTransactions(loopCtx context.Context, txm *SuiTxm, transactions []
 			// In the case there is an error submitting
 			txm.lggr.Errorw("Failed to broadcast transaction", "txID", tx.TransactionID, "function inputs", tx.Functions, "error", err)
 
-			var newState TransactionState
-			if resp.Effects.Status.Status == "" {
-				// If there is no status reported, we assume the transaction failed due to a network error or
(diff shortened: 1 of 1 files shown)


relayer/txm/broadcaster.go
```diff
@@ -88,14 +88,23 @@ func broadcastTransactions(loopCtx context.Context, txm *SuiTxm, transactions []
...
+				// An empty digest indicates a total failure of the transaction
 				txm.lggr.Errorw("Transaction failed without a digest", "txID", tx.TransactionID, "function inputs", tx.Functions)
-				err = txm.transactionRepository.ChangeState(tx.TransactionID, StateFailed)
-				if err != nil {
-					txm.lggr.Errorw("Failed to change transaction state to Failed", "txID", tx.TransactionID, "error", err)
-				}
+				newState = StateFailed
+			}
+
+			err = txm.transactionRepository.ChangeState(tx.TransactionID, newState)
```
FelixFan1992, line 102: newstate's default value is pending.
does it mean that if we have errors here but state is not "" and digest is not "", we still want to change state to pending?

faisal-chainlink: Good point, a better default here is likely `StateRetriable`

faisal-chainlink: This can been updated

FelixFan1992 (approved):

PR #247, opened 2025-11-12
update managed token publisher

FelixFan1992:


deployment/state.go
```diff
@@ -52,10 +52,11 @@ type CCIPChainState struct {
...
 	// Token related
-	ManagedToken                 string
-	ManagedTokenOwnerCapObjectID string
-	ManagedTokenStateObjectID    string
-	ManagedTokenMinterCapID      string
+	ManagedToken                  string
+	ManagedTokenOwnerCapObjectID  string
+	ManagedTokenStateObjectID     string
+	ManagedTokenMinterCapID       string
+	ManagedTokenPublisherObjectId string
```
rodrigombsoares: We'll need multiple managed tokens. This will have to be a slice or a map.

deployment/changesets/cs_deploy_managed_token.go
```diff
@@ -76,6 +76,13 @@ func (d DeployManagedToken) Apply(e cldf.Environment, config DeployManagedTokenC
 		return cldf.ChangesetOutput{}, fmt.Errorf("failed to save ManagedToken StateObjectId address %s for Sui chain %d: %w", managedTokenReport.Output.Objects.StateObjectId, config.ChainSelector, err)
 	}
 
+	// save PublisherObjectId address to the addressbook
+	typeAndVersionPublisherObjectId := cldf.NewTypeAndVersion(deployment.SuiManagedTokenPublisherObjectId, deployment.Version1_0_0)
```
rodrigombsoares, line 95: Recommend adding token symbol as a label here

deployment/state.go
FelixFan1992: done

deployment/changesets/cs_deploy_managed_token.go
FelixFan1992: updated.

deployment/state.go
```diff
@@ -51,11 +59,13 @@ type CCIPChainState struct {
...
-	ManagedTokenOwnerCapObjectID string
-	ManagedTokenStateObjectID    string
-	ManagedTokenMinterCapID      string
+	// Managed Token related
+	ManagedTokens map[string]ManagedTokenState
+	// ManagedToken                  string
+	// ManagedTokenOwnerCapObjectID  string
+	// ManagedTokenStateObjectID     string
+	// ManagedTokenMinterCapID       string
+	// ManagedTokenPublisherObjectId string
```
rodrigombsoares: we can delete this

FelixFan1992: done

deployment/state.go
```diff
@@ -166,15 +176,47 @@ func loadsuiChainStateFromAddresses(addresses map[string]cldf.TypeAndVersion) (C
...
 
-		// Token related
+		// Managed Token related
 		case SuiManagedTokenType:
-			chainState.ManagedToken = addr
+			symbol, err := getTokenSymbol(typeAndVersion)
+			if err != nil {
+				return CCIPChainState{}, fmt.Errorf("failed to get token symbol for Managed token: %w", err)
+			}
+			managed_token := chainState.ManagedTokens[symbol]
```
rodrigombsoares, line 181: You're gonna wanna instantiate this map explicitly, otherwise it's gonna be nil and we'll get `panic: assignment to entry in nil map`
Could you add these token types to the test case in `state_test.go` just so that we have this covered?

rodrigombsoares (changes requested):
Possible nil map panic

deployment/state.go
rodrigombsoares: You can do that directly on `chainState := CCIPChainState{` in the begging of the function

FelixFan1992:
> Possible nil map panic

good catch

rodrigombsoares (approved):

faisal-chainlink (approved):

PR #257, opened 2025-11-20
Fix: use hex for transaction digest in events

faisal-chainlink:
The E2E tests for this are expected to fail until core is updated to handle the hex tx digest instead of bytes.


FelixFan1992:
> The E2E tests for this are expected to fail until core is updated to handle the hex tx digest instead of bytes.

do you need to update only e2e tests in core or you need to update some actual logic in core?

faisal-chainlink:
> > The E2E tests for this are expected to fail until core is updated to handle the hex tx digest instead of bytes.
> 
> do you need to update only e2e tests in core or you need to update some actual logic in core?

The actual core logic. Primarily just the part that reads the tx digest part of the events to state machine changes / transaction state signals in core.

FelixFan1992 (approved):

FelixFan1992 (dismissed):

stackman27 (approved):

PR #265, opened 2025-11-23
Add pagination to `ReadFilterObjectIds` to enable fetching all matching objects

faisal-chainlink:


relayer/client/ptb_client.go
```diff
@@ -80,7 +80,7 @@ type SuiPTBClient interface {
 	MoveCall(ctx context.Context, req MoveCallRequest) (TxnMetaData, error)
 	SendTransaction(ctx context.Context, payload TransactionBlockRequest) (SuiTransactionBlockResponse, error)
 	ReadOwnedObjects(ctx context.Context, ownerAddress string, cursor *models.ObjectId) ([]models.SuiObjectResponse, error)
-	ReadFilterOwnedObjectIds(ctx context.Context, ownerAddress string, structType string, limit *uint) ([]models.SuiObjectData, error)
+	ReadFilterOwnedObjectIds(ctx context.Context, ownerAddress string, structType string, cursor string) ([]models.SuiObjectData, error)
```
FelixFan1992, line 84: if we remove the limit, what is the defualt limit?

faisal-chainlink: The behavior now is that it will always get all the owned objects which match the filter, it will paginate until its done

faisal-chainlink: Default limit is 50 per page for pagination

JohnChangUK (approved):

PR #275, opened 2025-11-26
chore(deployment): refactor mcms proposal building

rodrigombsoares:
core-ref:abb8603e3990f2f863502ca9916700add8869328

# Summary
Refactors MCMS proposal generation and its helpers into a `utils` package. This is the first step in the works of enabling regular changesets to work with MCMS.


RodrigoAD (approved):
👍

PR #281, opened 2025-12-01
Enhancement: Update transaction indexer to enable finding `ccip_receive` failures

faisal-chainlink:


relayer/chainreader/indexer/transactions_indexer.go
```diff
@@ -78,7 +78,7 @@ func NewTransactionsIndexer(
 		executionEventKey:       "ExecutionStateChanged",
 		configEventModuleKey:    "ocr3_base",
 		configEventKey:          "ConfigSet",
-		executeFunctions:        []string{"finish_execute"},
+		executeFunctions:        []string{"init_execute", "ccip_receive", "finish_execute"},
```
RodrigoAD: We should include token transfer processing, is it `release_or_mint`?

contracts/test/sources/offramp.move
```diff
@@ -450,6 +450,12 @@ module test::offramp {
         };
         event::emit(config_set);
 
+        let config_set2 = SourceChainConfigSet {
```
RodrigoAD, line 453: What is this for? Is it used?

RodrigoAD (approved):

contracts/test/sources/offramp.move
faisal-chainlink: This is needed because of the static report strings in the associated test.

PR #285, opened 2025-12-02
Minor: Event indexer thread safety and remove unused function

faisal-chainlink:


relayer/chainreader/indexer/events_indexer.go
kirk-baird, line 236: The lock should also be released before this return otherwise it could be poisoned.

faisal-chainlink: Addressed, thank you

FelixFan1992 (approved):

PR #291, opened 2025-12-04
Misc: TXM and Confirmer Refactor & Error Handling

faisal-chainlink:


relayer/txm/confirmer.go
```diff
@@ -92,126 +55,134 @@ func checkConfirmations(loopCtx context.Context, txm *SuiTxm) {
...
-			if err != nil {
-				txm.lggr.Errorw("Failed to update transaction gas", "transactionID", tx.TransactionID, "error", err)
-				return err
-			}
-			err = txm.transactionRepository.IncrementAttempts(tx.TransactionID)
-			if err != nil {
-				txm.lggr.Errorw("Failed to increment transaction attempts", "transactionID", tx.TransactionID, "error", err)
-				return nil
-			}
+	txm.lggr.Infow("Transaction is retriable", "transactionID", tx.TransactionID, "strategy", strategy)
```
kirk-baird: Nit pick: `retriable` -> `retryable` to match `IsRetryable()` above

relayer/txm/confirmer.go
```diff
@@ -92,126 +55,134 @@ func checkConfirmations(loopCtx context.Context, txm *SuiTxm) {
...
+		txm.lggr.Errorw("Failed to update transaction state", "transactionID", tx.TransactionID, "error", err)
+		return err
+	}
+
+	txm.broadcastChannel <- tx.TransactionID
+	return nil
+}
+
+func handleExponentialBackoffRetry(txm *SuiTxm, tx SuiTx) error {
+	delaySeconds := float64(3) * math.Pow(2, float64(tx.Attempt))
```
kirk-baird: Curious about whether 3 is too high. Sui blocktimes are pretty quick, 1 second might be enough as a starting point and I'm guessing that'd wouldn't cause any rate limiting issues.

Assuming max retires is 4, 1 second start would do 1,2,4,8 rather than 3,6,12,24.

It's a trade-off here, between how long we want to wait and whether it's likely to become valid in time. 8 seconds might be a little low for the last chance, so if 1 second is the starting point then I'd suggest increasing the max retries.

FelixFan1992:
core ref: 769837ac2e7c2529a3ca8392671bb16f1cb50b26

FelixFan1992 (approved):

PR #298, opened 2025-12-09
Update transaction indexer

faisal-chainlink:


relayer/chainreader/indexer/transactions_indexer_test.go
```diff
@@ -1,3 +1,5 @@
+//go:build integration
```
FelixFan1992, line 1: why this?

faisal-chainlink: We're using the same tag in other tests as well, put it here for consistency. This test was previously disabled and didn't include any tag.

faisal-chainlink: But more importantly, why not 😄

FelixFan1992 (approved):

PR #303, opened 2025-12-22
[fix] Handle LockCoin Error during txm Broadcast

stackman27:


relayer/chainreader/indexer/events_indexer_test.go
```diff
@@ -223,7 +223,7 @@ func TestEventsIndexer(t *testing.T) {
 				require.True(t, ok, "Event should have newValue field")
 
 				// The newValue should be i+1 (counter starts from 0, so increment makes it 1, 2, 3)
-				expectedValue := strconv.Itoa(i + 1)
+				expectedValue := strconv.Itoa(3 - i)
```
stackman27, line 226: imported from `develop` for test to pass

relayer/txm/transaction.go
```diff
@@ -91,11 +91,16 @@ func (tx *SuiTx) UpdateBSCPayload(
...
 	}
 
-	txBytes, paymentCoins, err := preparePTBTransaction(ctx, signerAddress, suiClient, tx.Ptb, tx.GasBudget, lggr)
+	// Reuse existing PTB + gas coins
+	tx.Ptb.SetGasBudget(tx.GasBudget)
+	tx.Ptb.SetSender(models.SuiAddress(signerAddress))
+	tx.Ptb.SetGasOwner(models.SuiAddress(signerAddress))
+	tx.Ptb.SetGasPayment(tx.PaymentCoinsObjectRef)
+
+	txBytes, err := toBCSBase64(ctx, tx.Ptb, signerAddress, lggr, tx.GasBudget)
```
faisal-chainlink, line 100: Curious why not update the `preparePTBTransaction` function instead?

relayer/txm/transaction.go
```diff
@@ -358,13 +363,36 @@ func preparePTBTransaction(
...
+					"coinObjectId", c.CoinObjectId,
+					"version", c.Version,
+					"key", key,
+				)
+				continue
+			}
+			filtered = append(filtered, c)
+		}
+		coinData = filtered
+	}
```
faisal-chainlink, line 394: I'm thinking it's better to take an approach where equivocation is prevented rather than handled after the lock occurs.

Given that by the time we receive an equivocation error, the coins are already locked, it's likely better to take a proactive approach. Thoughts?

faisal-chainlink (commented):
The PR currently attempts to handle an already equivocated object which is good since we cannot depend on the `LockedUntilEpoch` field, however, I'd recommend we take the approach of preventing equivocation and leveraging the exponential backoff strategy in cases for multiple transactions are attempted in-flight.

faisal-chainlink:
As an unrelated side note, the default `txm.configuration.TransactionRetentionSecs` is current 10 seconds which is too low. Leaving this comment here to discuss later.

faisal-chainlink:
Below are some notes and a simpler approach that I recommend.

Code context:
* The latest version of relayer codebase includes [categorization of errors](https://github.com/smartcontractkit/chainlink-sui/blob/develop/relayer/client/suierrors/errors.go#L253) into either gas errors or potentially re-triable errors.
* We have also added a default [exponential backoff retry](https://github.com/smartcontractkit/chainlink-sui/blob/develop/relayer/txm/confirmer.go#L110) strategy for errors which are not related to gas.
* The `confirmer` already picks up those inflight transactions (submitted or retriable) to handle them accordingly.
* Transactions that used the same coin object with the same version (i.e. concurrent transactions) should fail their broadcast round. 

The real problem:
* The broadcaster currently does not update the transaction object to include the error when setting its state to `Retriable`. This stop the `confirmer` from applying the exponential backoff to it and refreshing the coin objects.
* Exponential backoff re-queues the transaction without updating the gas coins' versions.

Simplest solution:
* Ensure that the relevant errors are categorized for exponential backoff by adding [those errors here](https://github.com/smartcontractkit/chainlink-sui/blob/develop/relayer/client/suierrors/errors.go#L245).
* Ensure that the `broadcaster` attaches the error when setting the transaction state to `Retriable`.
* Ensure that the `confirmer` updates gas coin versions for exponential backoff handling. This is currently only [done for gas bumps](https://github.com/smartcontractkit/chainlink-sui/blob/develop/relayer/txm/confirmer.go#L136).

stackman27:
hey @faisal-chainlink thanks for the feedback. Few thoughts; 

the generic retry pipeline you’re describing makes sense for a lot of error types, but for the Sui locked-coin / equivocation case specifically it doesn’t quite work by itself.

the locked-coin error is raised at broadcast time, inside SendTransaction in the broadcaster. So this class of error never becomes an “inflight” Submitted tx that the confirmer later sees. It dies right in broadcast. For confirmer to even see this error there has it has to be in SubmittedState, which is not the case here.

We can't go from pending -> retriable. We can try pending -> failed/submitted -> retriable, but this will require some hacks around txm as we're basically deceiving `GetTxStatus` call everytime during retries 

err below 
```
{"level":"error","ts":"2025-12-14T19:22:12.958Z","logger":"Sui.1.RelayerService.PluginRelayerClient.PluginSui.SuiRelayer.SuiTxm","caller":"loop/logger.go:164","msg":"Failed to change transaction state","version":"2.30.1@07443c6","chain":"sui","txID":"OffRamp-0xb9cb33ec10384413680e8327a1bd2c09704852b4132afa42075b90cdceac4b6c-6157b991-4988-4271-92d5-7eb48a6e3614","error":"pending state must transition to submitted or failed","stacktrace":"github.com/smartcontractkit/chainlink-sui/relayer/txm.broadcastTransactions\n\t/go/pkg/mod/github.com/smartcontractkit/chainlink-sui@v0.0.0-20251117202410-65d1d60c95af/relayer/txm/broadcaster.go:104\ngithub.com/smartcontractkit/chainlink-sui/relayer/txm.(*SuiTxm).broadcastLoop\n\t/go/pkg/mod/github.com/smartcontractkit/chainlink-sui@v0.0.0-20251117202410-65d1d60c95af/relayer/txm/broadcaster.go:54","caller":"txm/broadcaster.go:104","chainID":"1","timestamp":"2025-12-14T19:22:12.956Z","...} 
```

> Why exponential backoff + “refresh coin versions” isn’t enough

- The version only changes if some tx using that coin actually finalizes.
- If our tx never gets in (because Sui rejected it with this error), our local view of (objectID, version) doesn’t change.
- If we just: Mark the tx as Retriable, Later have the confirmer requeue it, And “refresh coin versions” without excluding that locked (objectID, version). This will resolve the existing stuck tx but the issue persist for upcoming txs right? 

We’re very likely to pick the same coin/version again and hit the same equivocation error, just later and with exponential backoff.

This PR ensures new txs aren't stuck from the nodes but doesnot address processing existing stuck tx (currently we just fail & let manual exec handle it), with confirmer retries we will be able to process the same stuck tx as well. But this seems like it'll need a slightly bigger txState refactor

faisal-chainlink:
> It dies right in broadcast. For confirmer to even see this error there has it has to be in SubmittedState, which is not the case here.

The confirmer also sees transactions in the retriable state. We can easily allow going from pending to retriable by simply adding a [key here](https://github.com/smartcontractkit/chainlink-sui/blob/develop/relayer/txm/store.go#L161) to allow this transition in the state machine.

> And “refresh coin versions” without excluding that locked (objectID, version). This will resolve the existing stuck tx but the issue persist for upcoming txs right?

It should eventually succeed as the retry will happen at a time when the previous transaction locking the coin object has already completed.

Remember that these transactions are processed sequentially and we've got the `WaitForEffectsCert`. We're essentially just handling an edge case here.

Another thing we can add to make this even safer is to refresh the coins at the time when we are broadcasting the transaction rather than at the time when we are queuing it.

> This PR ensures new txs aren't stuck from the nodes but doesnot address processing existing stuck tx (currently we just fail & let manual exec handle it)

While this PR is useful, I think it won't actually stop the problem from happening. This is because it doesn't prevent the actual use of a coin concurrently but rather keeps track of coins which already ran into the error of having a mismatched version. We still use a reactive approach instead of proactive one in this PR.

> But this seems like it'll need a slightly bigger txState refactor

Heads up that the transaction state machine has already been refactored in more recent commits in the `develop` branch.

stackman27:
closing in favor of https://github.com/smartcontractkit/chainlink-sui/pull/304

PR #304, opened 2025-12-25
Feat: Add a coin reservation abstraction to prevent equivocation within the same relayer

faisal-chainlink:
The PR includes the following: 

* Add a `GasCoinManager` abstraction to keep a cache of "reserved" coins (being used in other transacitons)
    * Current default expiry of reservation is 30 seconds
* Bring over some of the code in this [PR](https://github.com/smartcontractkit/chainlink-sui/pull/303) to include the LockedCoin error label and regex matching
* Add broadcast error parsing

Additional description of the change 

For a tx we build/send, you reserve its payment coins by calling `TryReserveCoins(txID, []SuiObjectRef, expiry=nil)` (or similar). That creates:

- `coinID -> true` with TTL = 30s (default)
- `txID -> []SuiObjectRef` with TTL = 30s

If we then detect the “locked coin / equivocation” issue and run the quarantine path, you call `TryReserveCoins(..., &expiry)` with 24h, which creates:

- `coinID -> true` with TTL = 24h
- `txID -> []SuiObjectRef` with TTL = 30s (still)

So after ~30s or failure/success:
- the txID mapping expires
- the coinID reservation remains for up to 24h

That’s why later ReleaseCoins(txID) often can’t undo the 24h quarantine: it loses the txID→coins index, but the coin lock key is still present.

### DIFF
--- a/relayer/client/suierrors/errors.go
+++ b/relayer/client/suierrors/errors.go
@@ -13,6 +13,8 @@ determine if an error is considered retryable (e.g. IsRetryable) according to Su
 package suierrors
 
 import (
+	"regexp"
+	"strconv"
 	"strings"
 )
 
@@ -28,6 +30,7 @@ const (
 	CheckpointAndConsensusErrors
 	PublishingErrors
 	SoftBundleErrors
+	LockCoinErrors
 	UnknownErrors
 )
 
@@ -47,6 +50,8 @@ func (c ErrorCategory) String() string {
 		return "Publishing Errors"
 	case SoftBundleErrors:
 		return "Soft Bundle Errors"
+	case LockCoinErrors:
+		return "Lock Coins Errors"
 	default:
 		return "Unknown Error Category"
 	}
@@ -146,6 +151,9 @@ var ErrNoSharedObjectError = NewSuiError(SoftBundleErrors, "NoSharedObjectError"
 var ErrAlreadyExecutedError = NewSuiError(SoftBundleErrors, "AlreadyExecutedError")
 var ErrCertificateAlreadyProcessed = NewSuiError(SoftBundleErrors, "CertificateAlreadyProcessed")
 
+// Lock Coins error
+var ErrLockCoins = NewSuiError(LockCoinErrors, "already locked by a different transaction")
+
 // Unknown Error
 var ErrUnknownError = NewSuiError(UnknownErrors, "UnknownError")
 
@@ -176,6 +184,7 @@ var suiErrorMappings = []struct {
 	{ErrUnsupported.Error(), ErrUnsupported},
 	{ErrMoveFunctionInputError.Error(), ErrMoveFunctionInputError},
 	{ErrPostRandomCommandRestrictions.Error(), ErrPostRandomCommandRestrictions},
+	{ErrObjectVersionUnavailableForConsumption.Error(), ErrObjectVersionUnavailableForConsumption},
 
 	// Gas Errors
 	{ErrMissingGasPayment.Error(), ErrMissingGasPayment},
@@ -225,9 +234,16 @@ var suiErrorMappings = []struct {
 	{ErrAlreadyExecutedError.Error(), ErrAlreadyExecutedError},
 	{ErrCertificateAlreadyProcessed.Error(), ErrCertificateAlreadyProcessed},
 
+	// Lock Coins error
+	{ErrLockCoins.Error(), ErrLockCoins},
+
 	// Unknown Error
 	{ErrUnknownError.Error(), ErrUnknownError},
 }
+
+var lockedObjectRe = regexp.MustCompile(
+	`Object\s+\((0x[0-9a-fA-F]+),\s*SequenceNumber\((\d+)\)`,
+)
 
 // ParseSuiErrorMessage maps a raw RPC error message to a structured error.
 // It iterates over the known substrings in suiErrorMappings. If a substring is found,
@@ -240,6 +256,25 @@ func ParseSuiErrorMessage(msg string) *SuiError {
 	}
 
 	return NewSuiError(UnknownErrors, msg)
+}
+
+// ExtractLockedObjectRef parses a Sui equivocation / lock error message and returns
+// (objectID, version, ok).
+func ExtractLockedObjectRef(msg string) (string, uint64, bool) {
+	m := lockedObjectRe.FindStringSubmatch(msg)
+	if len(m) != 3 {
+		return "", 0, false
+	}
+
+	objID := m[1]
+	verStr := m[2]
+
+	ver, err := strconv.ParseUint(verStr, 10, 64)
+	if err != nil {
+		return objID, 0, false
+	}
+
+	return objID, ver, true
 }
 
 var ExponentialBackoffErrors = []error{
--- a/relayer/client/suierrors/errors.go
+++ b/relayer/client/suierrors/errors.go
@@ -297,3 +297,7 @@ var GasBumpErrors = []error{
 	ErrInsufficientGas,
 	ErrGasPriceTooHigh,
 }
+
+var CoinRefreshErrors = []error{
+	ErrLockCoins,
+}
--- a/relayer/txm/confirmer.go
+++ b/relayer/txm/confirmer.go
@@ -163,6 +163,8 @@ func handleTransactionError(ctx context.Context, txm *SuiTxm, tx SuiTx, result *
 		return handleExponentialBackoffRetry(txm, tx)
 	case GasBump:
 		return handleGasBumpRetry(ctx, txm, tx, txError)
+	case CoinRefresh:
+		return handleCoinRefreshRetry(ctx, txm, tx, txError)
 	case NoRetry:
 		return markTransactionFailed(txm, tx, txError)
 	default:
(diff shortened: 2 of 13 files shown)


relayer/client/suierrors/errors.go
faisal-chainlink, line 1: Changes to this file are brought over from #303

relayer/txm/confirmer.go
```diff
@@ -95,8 +109,40 @@ func handleTransactionError(ctx context.Context, txm *SuiTxm, tx SuiTx, result *
...
+		coinID, err := transaction.ConvertSuiAddressStringToBytes(models.SuiAddress(objectID))
+		if err == nil && !txm.coinManager.IsCoinReserved(*coinID) {
+			// Coin lock duration
+			expiry := 24 * time.Hour
+			
+			// The coin is not recorded is not marked as reserved, mark it as reserved
+			err = txm.coinManager.TryReserveCoins(ctx, tx.TransactionID, []transaction.SuiObjectRef{
+				{
+					ObjectId: *coinID,
+					Version:  0,
```
stackman27, line 137: can we lock the version too?

relayer/txm/confirmer.go
```diff
@@ -95,8 +109,40 @@ func handleTransactionError(ctx context.Context, txm *SuiTxm, tx SuiTx, result *
...
+		txm.lggr.Infow("Detected locked coin at confirmation time",
+			"txID", tx.TransactionID,
+			"objectID", objectID,
+			"version", version,
+		)
+		
+		coinID, err := transaction.ConvertSuiAddressStringToBytes(models.SuiAddress(objectID))
+		if err == nil && !txm.coinManager.IsCoinReserved(*coinID) {
+			// Coin lock duration
+			expiry := 24 * time.Hour
```
stackman27: nit: `DefaultLockedCoinTTL`

relayer/txm/confirmer.go
```diff
@@ -56,14 +58,26 @@ func checkConfirmations(loopCtx context.Context, txm *SuiTxm) {
...
+			txm.lggr.Debugw("Transaction is in submitted state", "transactionID", tx.TransactionID)
+			resp, err = txm.suiGateway.GetTransactionStatus(loopCtx, tx.Digest)
+			if err != nil {
+				txm.lggr.Errorw("Error getting transaction status", "transactionID", tx.TransactionID, "error", err)
+				continue
+			}
+		} else if tx.State == StateRetriable {
+			txm.lggr.Debugw("Transaction is in retriable state", "transactionID", tx.TransactionID)
+			// Check if it's a broadcast error (never made it onchain)
+			if tx.BroadcastError == "" {
```
stackman27, line 75: the broadcast error tx never makes it onchain but the rpc does return an error which we seem to store it in internal txm memory here (https://github.com/smartcontractkit/chainlink-sui/pull/304/changes#diff-31590b36619a1e616923716f8994038ae7cf64a025a040d289de96434ad81ce5R97) 

thinking of a flow where we would hit this case at all 🤔

relayer/txm/gas_coin_manager.go
```diff
@@ -0,0 +1,90 @@
...
+	"github.com/block-vision/sui-go-sdk/transaction"
+	"github.com/patrickmn/go-cache"
+	"github.com/smartcontractkit/chainlink-common/pkg/logger"
+
+	"github.com/smartcontractkit/chainlink-sui/relayer/client"
+)
+
+const (
+	DefaultLockedCoinTTL = 24 * time.Hour
+	DefaultAllocationTimeout = 30 * time.Second
```
stackman27, line 19: Ideally want to reduce this as much as possible (maybe by half or more) since txs are processed so fast in SUI in ideal case 

Commit Plugin sends multiple txs per minute, can we have a shorter window to prevent backfill overhead and potential race conditions?

relayer/txm/gas_coin_manager.go
```diff
@@ -0,0 +1,90 @@
...
+	for _, coin := range coinIDs.([]transaction.SuiObjectRef) {
+		coinID := hex.EncodeToString(coin.ObjectId[:])
+		m.coinsCache.Delete(coinID)
+	}
+
+	m.coinsCache.Delete(txID)
+	return nil
+}
+
+func (m *SuiGasCoinManager) IsCoinReserved(coinID models.SuiAddressBytes) bool {
```
stackman27, line 88: slight preference, can we store version as well just so we are aware at which version it's locked and what version to expect once it's unlocked

relayer/txm/confirmer.go
stackman27, line 161: we probably need to add the LockError error under one of the following strategy right? Also we still need to add selection of new coin for retry since we're not rebuilding the tx via EnquePTB

relayer/txm/txm.go
```diff
@@ -105,6 +109,12 @@ func (txm *SuiTxm) EnqueuePTB(ctx context.Context, transactionID string, txMetad
 	txm.lggr.Infow("PTB Transaction added to broadcast channel", "transactionID", transactionID)
 	txm.lggr.Infow("PTB Transaction enqueued", "transactionID", transactionID)
 
+	err = txm.coinManager.TryReserveCoins(ctx, transactionID, txn.PaymentCoinsObjectRef, nil)
```
stackman27, line 112: this logic runs regardless of broadcast failure or success right? Will we have a race condition where it'll overwrite an already locked coin (from 24hrs -> 30sec?)

relayer/txm/confirmer.go
faisal-chainlink: We can but it doesn't really make any difference here since we're primarily interesting in the object's ID when it comes to hashing a key for the cache entry.

relayer/txm/confirmer.go
faisal-chainlink: Yea a good example of this is the RPC rejecting the transaction immediately without being acceptable as a valid transaction. Input errors / mistakes are a common case for this.

relayer/txm/gas_coin_manager.go
faisal-chainlink: Yea we can definitely safely reduce this further. An even better approach is to keep this as is and call the release method whenever we confirm a transaction is successful or failed in the confirmer. This would essentially keep the 30 seconds TTL as a fallback.

relayer/txm/confirmer.go
faisal-chainlink: The re-selection should be taken care of in the save flow of rebuilding a transaction via `tx.UpdateBSCPayload` (invoked in `UpdateTransactionGas()` which is currently only called within the GasBump retry logic). However, we need to make a decision on whether those transactions with no available coins to use (essentially can be classified as "InsufficientGas") should fail and emit that to the core node or retry internally with the relayer.

relayer/txm/txm.go
faisal-chainlink: This logic is pre-broadcast entirely. This method responds back to the core node to tell it if the submission attempt has been queued for broadcasting. We don't go from 24 hours -> 30 seconds in any of the flows. 

Here are a few scenarios for this LoC:

Scenario 1: There is a coin which is not reserved and we attempt to reserve it. This call will succeed.

Scenario 2: All valid coins are reserved. In this case the enqueuing won't make this far. It will fail at the `GeneratePTBTransactionWithGasEstimation()` call earlier in this file as there won't be any coins usable for gas which are not reserved or have sufficient balance. This will respond back to the core node's `SubmitTransaction` call with a failure.

I'm leaning towards this approach as it will immediately inform the core node that the submission has failed and then the core node can actually re-attempt the submission very quickly. This is a different flow that a transaction being submitted successfully but later being marked as failed with an update `ExecutionStateChanged` event.

relayer/txm/gas_coin_manager.go
faisal-chainlink: We can add it. Although I'm thinking of simply using the available SDK client for fetching these details when its needed. Currently the locking is not version dependent as equivocation takes place when the concurrent transactions are using the same version. 

The mismatched version should yield a separate error which currently goes through the exponential backoff. For example: `ErrObjectVersionUnavailableForConsumption`.

relayer/txm/confirmer.go
stackman27: > essentially can be classified as "InsufficientGas"

i don't think we should classify it as insufficientGas, i'm in favor of a separate error type for it

relayer/txm/gas_coin_manager.go
stackman27: yea that should be fine to

relayer/txm/gas_coin_manager.go
stackman27: yea in favor of instant release after successful tx & using TTL as fallback

relayer/txm/txm.go
stackman27: ah i see `IsCoinReserved` so there is indeed a fail safe 

in the case of Scenario 2, we can assume it'll never reach broadcast right?

relayer/txm/confirmer.go
faisal-chainlink: It is a separate error type but I meant the error category in terms of a retry strategy.

relayer/txm/gas_coin_manager.go
faisal-chainlink: Added instant release on success or failure

relayer/txm/txm.go
faisal-chainlink: > we can assume it'll never reach broadcast right?

Yes, that's right

relayer/txm/confirmer.go
```diff
@@ -147,6 +201,40 @@ func handleGasBumpRetry(ctx context.Context, txm *SuiTxm, tx SuiTx, txError *sui
 	return nil
 }
 
+func handleCoinRefreshRetry(ctx context.Context, txm *SuiTxm, tx SuiTx, txError *suierrors.SuiError) error {
+	txm.lggr.Infow("Coin refresh strategy - refreshing coins for locked coin error", "transactionID", tx.TransactionID)
+
+	// Release the old coins that are locked
+	if err := txm.coinManager.ReleaseCoins(tx.TransactionID); err != nil {
```
stackman27, line 208: do we want to release the coins here? we just reserved the coins during `handleTransactionError` 

can we auto release it in 24hrs via cacheExpiry?

stackman27: nvm i see there's 2 different cache keys

stackman27 (approved):
LGTM

FelixFan1992 (approved):

PR #306, opened 2026-01-05
Fix: Duplicate offsets in events indexer

faisal-chainlink:
What's in the PR?

* Revert the original events query without deduping
* Remove just-in-time events sync when calling `QueryKey` in the relayer
  * This will reduce the RPC load, make the responses quicker and remove the source of the events duplication
* Add a new DB index to use an `md5` hash over the event data along with the `tx_digest` and `event_handle` to ensure that duplicate events are not inserted.

> IMPORTANT: Since we are adding a unique index, existing duplicates will require a clean up query to ensure that they are removed. The index creation will fail otherwise.

### DIFF
--- a/relayer/chainreader/database/database.go
+++ b/relayer/chainreader/database/database.go
@@ -176,7 +176,19 @@ func (store *DBStore) GetLatestOffset(ctx context.Context, eventAccountAddress, 
 	var offset uint64
 	var txDigest string
 	var totalCount uint64
-	err := store.ds.QueryRowxContext(ctx, QueryEventsOffset, eventAccountAddress, eventHandle).Scan(&offset, &txDigest, &totalCount)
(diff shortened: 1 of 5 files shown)


faisal-chainlink:
For removing the existing duplicates, we can use the following query: 

```
WITH duplicates AS (
	SELECT id, ROW_NUMBER() OVER (
		PARTITION BY tx_digest, block_height, md5(data::text)
		ORDER BY event_offset DESC
	) as rn
	FROM sui.events
	WHERE event_account_address = $1 AND event_handle = $2
)
DELETE FROM sui.events
WHERE id IN (
	SELECT id FROM duplicates WHERE rn > 1
)
```

stackman27 (dismissed):

stackman27 (dismissed):

stackman27 (dismissed):

stackman27 (approved):

PR #318, opened 2026-01-12
use []any instead of []uint8 for messageId and messageHash in synth events

faisal-chainlink:


relayer/chainreader/indexer/transactions_indexer.go
```diff
@@ -491,11 +491,14 @@ func (tIndexer *TransactionsIndexer) syncTransmitterTransactions(ctx context.Con
 			executionStateChanged := map[string]any{
 				"source_chain_selector": fmt.Sprintf("%d", sourceChainSelector),
 				"sequence_number":       fmt.Sprintf("%d", execReport.Message.Header.SequenceNumber),
-				"message_id":            execReport.Message.Header.MessageID,
-				"message_hash":          messageHash[:],
-				"state":                 uint8(3), // 3 = FAILURE
+				// The conversion to []any is needed to avoid the default Go DB SDK behaviour of converting the byte slice to encoded base64 string.
+				"message_id":   codec.BytesToAnySlice(execReport.Message.Header.MessageID),
+				"message_hash": codec.BytesToAnySlice(messageHash[:]),
```
stackman27, line 496: should we decode these values first? 
```
base64.StdEncoding.DecodeString(messageId)
```

faisal-chainlink: They are already `[]byte` or `[32]byte` not strings. The issue here is that when inserting them as those types, the DB SDK auto-converts it to `base64`, all we're doing here is going from `[]byte` to `[]any` to match the JSON-friendly types that are inserted with regular events (comes from `parsedJson`).

This conversion effectively tells the DB SDK, we want to insert this slice as-is.

stackman27 (approved):

relayer/chainreader/indexer/transactions_indexer_test.go
```diff
@@ -450,6 +461,15 @@ func TestTransactionsIndexer(t *testing.T) {
...
 			return hasEvent(boundContracts[0], "ExecutionStateChanged")
 		}, 90*time.Second, 5*time.Second)
+
+		events, err := cReader.QueryKey(ctx, boundContracts[0], query.KeyFilter{Key: "ExecutionStateChanged"}, query.LimitAndSort{}, &OfframpExecutionStateChanged{})
+		require.NoError(t, err)
+		require.NotEmpty(t, events)
+
+		executionStateChanged := events[0].Data.(*OfframpExecutionStateChanged)
+
+		require.True(t, strings.HasPrefix(executionStateChanged.MessageId, "0x"))
```
RodrigoAD, line 471: Why not check the exact `messageId`? Same with `messageHash`

faisal-chainlink: While we could parse `reportStr` and decode the report it has to get the actual message ID. It's not necessary imo. 

The check just needs to verify that the type was read correctly. Given that it's a new contract being published, we can be fairly certain it's the same event we're looking for.

faisal-chainlink: Added the exact messageId value check here: https://github.com/smartcontractkit/chainlink-sui/pull/301

PR #325, opened 2026-01-16
Fix MCMS dynamic proposal generator YAML. List Ops script

RodrigoAD:
- Fixes the registry of operations to use `AsUntypedRelaxed()` so they can be used with YAML/JSON inputs
- Adds script to list every operation in the Op Registry available for MCMS, with its definition and YAML input template
```
(nix:nix-shell-env) chainlink-sui $ task utils:list-operations
All Operations with Input Templates:
====================================

...

89. ID: sui-ccip-rmn_remote-uncurse_chain
   Version: 0.1.0
   Description: Uncurse a chain selector in the CCIP RMN Remote contract
   Input Type: rmn.CurseUncurseChainInput
   YAML Template:
     ccippackageid: # string
     stateobjectid: # string
     ownercapobjectid: # string
     subjects:
       - # []uint8

Total operations: 89
```


scripts/list_operations.go
```diff
@@ -0,0 +1,178 @@
...
+				fmt.Print(yaml)
+				fmt.Println()
+			}
+		}
+	}
+
+	fmt.Printf("Total operations: %d\n", len(allOps))
+}
+
+func generateYAMLTemplate(t reflect.Type, indent string, depth int, visited map[reflect.Type]bool, maxDepth int) string {
```
RodrigoAD, line 59: Inspired by https://github.com/smartcontractkit/chainlink-deployments-framework/blob/f04042959eea559ab74123d1dbc6c29c2d70bba1/engine/cld/legacy/cli/commands/template_input.go#L126

deployment/ops/ccip/op_registry.go
```diff
@@ -1,47 +1,43 @@
...
-	*FeeQuoterInitializeOp.AsUntyped(),
-	*FeeQuoterApplyFeeTokenUpdatesOp.AsUntyped(),
-	*FeeQuoterApplyTokenTransferFeeConfigUpdatesOp.AsUntyped(),
-	*FeeQuoterApplyDestChainConfigUpdatesOp.AsUntyped(),
-	*FeeQuoterApplyPremiumMultiplierWeiPerEthUpdatesOp.AsUntyped(),
-	*FeeQuoterUpdateTokenPricesOp.AsUntyped(),
-	*FeeQuoterNewFeeQuoterCapOp.AsUntyped(),
-	*FeeQuoterDestroyFeeQuoterCapOp.AsUntyped(),
-	*FeeQuoterUpdatePricesWithOwnerCapOp.AsUntyped(),
+	*FeeQuoterInitializeOp,
```
FelixFan1992, line 6: in order to make dynamic changeset working, i followed Graham's suggestion to use AsUntypedRelaxed for Aptos: https://github.com/smartcontractkit/chainlink/pull/20688/changes#diff-a5f92d5f2785465be318993b555db77a8a6ef61c4a001467b604f007f161994cR21

do you think it's better to do this?

RodrigoAD: Yeah exactly, I did that here: https://github.com/smartcontractkit/chainlink-sui/blob/fix-mcms-dynamic-yaml/deployment/ops/registry/op_registry.go#L64

I'm exporting them now with the types to make the script work, and then exporting them with `UnTypedRelaxed` for CLD

deployment/ops/registry/op_registry.go
```diff
@@ -34,8 +36,43 @@ var AllOperations = func() []cld_ops.Operation[any, any, any] {
...
 	// RMN Operations
 	operations = append(operations, rmnops.AllOperationsRMN...)
-	// Add more operation slices here as needed:
-	// operations = append(operations, anotherops.AllOperations...)
+
+	return operations
+}()
+
+// Exports every operation available so they can be registered to be used in dynamic changesets
+var AllOperations = func() []cld_ops.Operation[any, any, any] {
```
rodrigombsoares: Since we're already refactoring, can we return a list of pointer instead? Since `NewOperationRegistry` expects pointers
https://github.com/smartcontractkit/chainlink-deployments/blob/d17d1912b19ebf428f649455ae21cb81b2de0780/domains/ccip/testnet/durable_pipelines.go#L2179-L2184

deployment/ops/registry/op_registry.go
```diff
@@ -34,8 +36,43 @@ var AllOperations = func() []cld_ops.Operation[any, any, any] {
...
+			// If we can't get the address directly (e.g., because it's stored as an interface value),
+			// we need to create a new addressable value
+			opPtr := reflect.New(opVal.Type())
+			opPtr.Elem().Set(opVal)
+			opVal = opPtr
+		}
+
+		asUntypedMethod := opVal.MethodByName("AsUntypedRelaxed")
+		if !asUntypedMethod.IsValid() {
+			panic(fmt.Sprintf("operation %v does not have AsUntypedRelaxed method", opVal.Type()))
```
rodrigombsoares, line 66: This function is used in other repos, let's return error instead of panicking

deployment/ops/registry/op_registry.go
```diff
@@ -34,8 +36,43 @@ var AllOperations = func() []cld_ops.Operation[any, any, any] {
...
+		}
+		results := asUntypedMethod.Call(nil)
+		if len(results) != 1 {
+			panic("AsUntypedRelaxed should return exactly one value")
```
rodrigombsoares, line 70: ditto

deployment/ops/registry/op_registry.go
```diff
@@ -14,9 +17,8 @@ import (
 	rmnops "github.com/smartcontractkit/chainlink-sui/deployment/ops/rmn"
 )
 
-// Exports every operation available so they can be registered to be used in dynamic changesets
-var AllOperations = func() []cld_ops.Operation[any, any, any] {
-	var operations []cld_ops.Operation[any, any, any]
+var AllOperationsTyped = func() []any {
```
rodrigombsoares, line 20: Is there a usecase for this `AllOperationsTyped: []any`? If not, can we just pass operations as untype relaxed at once to simplify? What is the advantage of having this `AllOperationsTyped` function?

RodrigoAD: If we want to group them as `cld_ops.Operation[any, any, any]`, we lose the generics using the `AsUntypedRelaxed`, so the script to generate the template can't access the input type. Having them as `any` allow this, although creates a bit of complexity using `reflect`

deployment/ops/registry/op_registry.go
```diff
@@ -34,8 +36,43 @@ var AllOperations = func() []cld_ops.Operation[any, any, any] {
...
+
+	operations := make([]cld_ops.Operation[any, any, any], len(typedOps))
+	for i, op := range typedOps {
+		// Use reflection to call AsUntyped method
+		// The operations are stored as values, so we need to get their address to call pointer methods
+		opVal := reflect.ValueOf(op)
+
+		// If AsUntyped is a pointer receiver, we need to get the address
+		if opVal.CanAddr() {
+			opVal = opVal.Addr()
```
rodrigombsoares, line 55: Are we ever going to reach this branch?

deployment/ops/registry/op_registry.go
RodrigoAD: This is evaluated at compilation, not at runtime, so is very unlikely that this ever panics to downstream users, will happen to us first

deployment/ops/registry/op_registry.go
RodrigoAD: Yeah true, let me simplify it 

edit: we actually need it, if we keep each registry exporting the Op pointers

FelixFan1992 (approved):

PR #328, opened 2026-01-16
update SUI CLI to sui 1.65.2

FelixFan1992:


contracts/mcms/mcms/tests/mcms_test.move
```diff
@@ -172,97 +172,97 @@ fun default_execute_args(): ExecuteArgs {
     }
 }
 
-#[test]
```
FelixFan1992, line 175: currently, we can use a hack:

```
[published.testnet]
chain-id = "4c78adac"
published-at = "0x0000000000000000000000000000000000000000000000000000000000000000"
original-id = "0x0000000000000000000000000000000000000000000000000000000000000000"
version = 1
```

but this is not ideal. waiting for Sui team to release a new approach

sui.nix
```diff
@@ -9,13 +9,13 @@ stdenv.mkDerivation rec {
   src = if stdenv.hostPlatform.isDarwin then
     pkgs.fetchzip {
       url = "https://github.com/MystenLabs/sui/releases/download/mainnet-v${version}/sui-mainnet-v${version}-macos-arm64.tgz"; # Assume is a M1 Mac
-      sha256 = "sha256-9NniV7tc+4OGoPUiB9NO0K/Sy36TPnxhQqlW3ZAMyIM=";  # Should be replaced when bumping versions
+      sha256 = "sha256-L64Wj2RitsqVymra2sQgrSI3J2fzTtbBE+W7DWMW3zs=";  # Should be replaced when bumping versions
```
RodrigoAD: MacOS hash should be different than Linux, just run it locally and will tell you the expected one

faisal-chainlink:
Seems like the primary remaining item in this PR is the insufficient gas error. Likely due to account selection in the CLI.

faisal-chainlink:
@FelixFan1992 Currently all tests should pass with the exception of the upgrades via MCMS. The issue with MCMS upgrades is that we are still using the CLI the build the compiled package.

> Note: I'm running the relaying and binding tests with `go test ./... -v -tags=integration -count=1 -p=1` from the main directory. For the `/integration-tests` directory, it's the same command but needs to be run within the subdirectory since it's a separate module. This is slightly different from how CI is running them but we can handle that afterwards.

When using the `--with-unpublished-dependencies`, I'm seeing an "incompatible upgrade" error. However, if we use the regular publishing flow to generate the byte code of a package, we must also fill in all the addresses of the dependencies which isn't currently done.

That should be the last missing piece. 

---

The summary is that we are now creating the `Published.toml` file or `pub.[env].toml` file whenever we want to handle a given contract deployment. We start by traversing all the deps for that contract and build their `toml` files one by one to enable publishing / compiling.

faisal-chainlink:
Was able to get TokenPool upgrades via MCMS working. The remaining errors are the same in most test suites and are likely related to the setup of the test itself.

faisal-chainlink:
Note: There is one test in the onramp integration suite that I disabled in favor of using E2E tests because I think that test is missing a couple items causing runs to be flakey. In particular, the expected event isn't being captured.

.github/actions/setup-sui/action.yml
```diff
@@ -2,9 +2,9 @@ name: Setup Sui CLI
 description: Install and cache Sui CLI using suiup
 inputs:
   version:
-    description: Sui CLI version to install (e.g., mainnet-v1.61.2)
+    description: Sui CLI version to install (e.g., testnet-v1.64.0)
     required: false
-    default: mainnet-v1.61.2
+    default: testnet-v1.64.0
```
RodrigoAD: Why are we using testnet? Shouldn't it be `mainnet`?

.github/workflows/sui-ccip-test.yml
```diff
@@ -20,27 +20,27 @@ jobs:
...
-          - test_name: Test_CCIPTokenTransfer_Sui2EVM_BurnMintTokenPool_ThenGloballyCursedUncursed
-            sui_version: mainnet-v1.60.1
-          - test_name: Test_CCIPTokenTransfer_Sui2EVM_ManagedTokenPool_ThenCurseUncurse
-            sui_version: mainnet-v1.60.1
-          - test_name: Test_CCIPTokenTransfer_Sui2EVM_ManagedTokenPool_WithRateLimit
-            sui_version: mainnet-v1.60.1
-          - test_name: Test_CCIP_Upgrade_Sui2EVM
-            sui_version: mainnet-v1.60.1
+            sui_version: testnet-v1.64.0
+          # - test_name: Test_CCIP_EVM2Sui_ZeroReceiver
```
RodrigoAD: Why are all these commented out?

bindings/bind/bcs_decoder_test.go
```diff
@@ -331,3 +331,27 @@ func TestDeserializer(t *testing.T) {
 		des,
 	)
 }
+
+/**
+func TestDeserializer_ShouldFail(t *testing.T) {
+	// TODO: this test will fail when https://github.com/block-vision/sui-go-sdk/pull/78
+	// is released and will have to be removed
```
RodrigoAD: We can already remove it then, right?

bindings/bind/publish.go
```diff
@@ -57,11 +57,14 @@ func PublishPackage(
...
 	}
 
-	gasBudgetValueDefault := uint64(500_000_000)
+	gasBudgetValueDefault := uint64(1_000_000_000)
 	if opts.GasBudget == nil {
 		opts.GasBudget = &gasBudgetValueDefault // 500M MIST default for publish
 	}
 
+	fmt.Printf("ptb.Publish gasBudget: %d\n", *opts.GasBudget)
+	fmt.Printf("ptb.Publish dependencies: %+v\n", deps)
```
RodrigoAD:
```suggestion
```

bindings/bind/published_toml.go
```diff
@@ -0,0 +1,398 @@
...
+
+	// Normalize addresses to have 0x prefix
+	publishedAt = normalizeAddress(publishedAt)
+	originalID = normalizeAddress(originalID)
+
+	content := fmt.Sprintf(`# Generated by chainlink-sui bindings
+# This file contains metadata about published versions of this package
+# This file is auto-generated for dependency resolution
+
+[published.%s]
```
RodrigoAD, line 57: This file is a bit difficult to follow and it'll be harder to maintain over time. Let's please switch to a TOML library to handle the TOML parsing/serialization (e.g., go-toml: https://github.com/pelletier/go-toml) now that we have context on what it does 🙏

bindings/packages/test/test.go
```diff
@@ -78,19 +79,28 @@ func PublishTest(ctx context.Context, opts *bind.CallOpts, client sui.ISuiAPI, t
 		return nil, nil, err
 	}
 
+	// chainID, err := testutils.GetChainIdentifier(suiRPC)
+	// if err != nil {
+	// 	return nil, nil, err
+	// }
+
+	// testutils.PatchEnvironmentTOML("contracts/test", "local", chainID)
```
RodrigoAD:
```suggestion
```

bindings/packages/test/test.go
```diff
@@ -78,19 +79,28 @@ func PublishTest(ctx context.Context, opts *bind.CallOpts, client sui.ISuiAPI, t
...
+		"signer":         signerAddr,
 	}, false, suiRPC)
 	if err != nil {
 		return nil, nil, err
 	}
 
-	if err != nil {
-		return nil, nil, err
-	}
+	artifact.Dependencies = append(artifact.Dependencies, testSecondary)
```
RodrigoAD, line 91: This should be handled by the `CompilePackage` function

bindings/packages/test/test.go
```diff
@@ -78,19 +79,28 @@ func PublishTest(ctx context.Context, opts *bind.CallOpts, client sui.ISuiAPI, t
...
+
+	log.Printf("suiRPC: %s\n", suiRPC)
+	log.Printf("successfully compiled test package with test_secondary: %s\n", testSecondary)
+	log.Printf("modules: %+v\n", artifact.Modules)
+	log.Printf("dependencies: %+v\n", artifact.Dependencies)
```
RodrigoAD: I see some debugging logs in this test, let's please remove them to avoid noise
```suggestion
```

bindings/packages/test/test.go
RodrigoAD: Let's try to clean all these comments 👍 
```suggestion
```

bindings/tests/binding_features_test.go
```diff
@@ -28,9 +29,18 @@ func TestBindingFeatures(t *testing.T) {
 		GasBudget:        &DEFAULT_GAS_BUDGET,
 	}
 
+	chainID, err := client.SuiGetChainIdentifier(ctx)
+	if err != nil {
+		t.Fatalf("failed to get chain identifier: %v", err)
+	}
+
+	testutils.PatchEnvironmentTOML("contracts/test_secondary", "local", chainID)
```
RodrigoAD, line 37: What is this for?

contracts/ccip/ccip/Published.toml
```diff
@@ -0,0 +1,9 @@
+# Generated by Move
+# This file contains metadata about published versions of this package in different environments
+# This file SHOULD be committed to source control
+
+[published.testnet]
+chain-id = "4c78adac"
+published-at = "0x5ef4b483da6644c84aa78eae4f51a9bfb1fb4554d5134ac98892e931fcbdd6bf"
```
RodrigoAD, line 7: Are these real testnet addresses? If not, I'd avoid committing them or the file

contracts/ccip/ccip_token_pools/usdc_token_pool/Move.toml
```diff
@@ -1,34 +1,34 @@
...
-usdc_token_pool = "0x0"
-usdc_coin_metadata_object_id = "_"
-token_messenger_minter = "_"
-token_messenger_minter_state = "_"
-message_transmitter = "_"
-message_transmitter_state = "_"
-treasury = "_"
-mcms = "_"
-mcms_owner = "_"
+# [addresses]
```
RodrigoAD, line 12: We can clean all these commented part

bindings/bind/compile.go
```diff
@@ -223,6 +223,7 @@ func compilePackageInternal(packageName contracts.Package, namedAddresses map[st
 
 	// before you set the temp dir
 	prevConfigDir := os.Getenv("SUI_CONFIG_DIR")
+	log.Printf("Previous SUI_CONFIG_DIR: %s", prevConfigDir)
```
RodrigoAD: Let's please clean up all the development debugging logs

bindings/bind/published_toml.go
```diff
@@ -0,0 +1,398 @@
...
+	if err := os.WriteFile(filePath, []byte(sb.String()), 0o644); err != nil {
+		return fmt.Errorf("failed to append to pub file: %w", err)
+	}
+
+	return nil
+}
+
+// UpdatePublishedTOMLForUpgrade updates an existing Published.toml file for a package upgrade.
+// This updates the published-at address and increments the version.
+func UpdatePublishedTOMLForUpgrade(packageDir, environment, chainID, newPublishedAt, originalID string, newVersion int, upgradeCap string) error {
```
RodrigoAD, line 176: Are `UpdatePublishedTOMLForUpgrade`, `RemovePublishedTOML`, `WriteEphemeralPubFile` and `removePublishedEntry` used? If not, we can remove them

bindings/bind/compile.go
```diff
@@ -295,7 +358,7 @@ func compilePackageInternal(packageName contracts.Package, namedAddresses map[st
 		mcmsAddr := namedAddresses["mcms"]
 		if !isZeroAddress(mcmsAddr) {
 			mcmsDir := filepath.Join(dstRoot, "mcms", "mcms")
-			if err := managePackage(mcmsDir, 1, rpcURL, env, mcmsAddr, mcmsAddr); err != nil {
+			if err := managePackage(mcmsDir, 1, rpcURL, env, mcmsAddr, mcmsAddr, pubfilePath); err != nil {
```
RodrigoAD, line 345: `originalPkgId` and `latestPkgId` seem to be always the same for every contract. I can't recall why we had it different. Could we just unify them?

.github/actions/setup-sui/action.yml
FelixFan1992: yeah let's change this to v1.65.2

bindings/bind/bcs_decoder_test.go
FelixFan1992: yes this should be removed

contracts/ccip/ccip/Published.toml
FelixFan1992: this was a hack to make sure tests are passing [here](https://github.com/smartcontractkit/chainlink-sui/pull/328/changes#diff-b42562494330f19eda573f3113ef124a83e2ac3a22302197580416baedd83129R502)

we can try to remove this OR remove those test cases.

bindings/bind/compile.go
faisal-chainlink: We can probably unify those. I think we have a test for an upgrade to an already upgraded package in please. IIRC that might have been the reason we had it there but not entirely sure.

faisal-chainlink (approved):

PR #331, opened 2026-01-19
Enhancement: health reporting and metrics

faisal-chainlink:
This PR adds a health monitor abstraction to enable the various components of the relayer to report their health status via telemetry metrics.

It enables the following:
- Monitoring component status via Grafana
- Detecting unstable components via flip-flop signals
- Identifying stalling processors via lag metrics

NOPS can setup their alerts on the basis of these metrics.


makramkd (dismissed):

faisal-chainlink:
Seems that there's a change on the core repo that's causing the E2E tests to fail

faisal-chainlink:
Closed in favor of including the changes in #500

PR #336, opened 2026-01-28
Fix: handle offset edge case

faisal-chainlink:
The current `offset` stored for events takes into account the `EventSeq` read from the RPC along with the total count.

The issue is that this can cause a jump in the `offset` in specific scenarios. 

The resolution simply uses a value of `1` instead of `EventSeq` along with the `totalCount` to make sure it is always correctly sequential.


stackman27 (approved):
LGTM Thank you!

PR #348, opened 2026-03-12
Fastcurse tooling API adapter

RodrigoAD:
- Implements `fastcurse` for the CCIP tooling API
- Adds integration tests against local env for MCMS and non-MCMS workflows using Adapter


bindings/bind/compile.go
```diff
@@ -320,7 +320,6 @@ func compilePackageInternal(packageName contracts.Package, namedAddresses map[st
 		// Write Published.toml for test_secondary dependency if its address is provided
 		testSecondaryAddr := namedAddresses["test_secondary"]
 		if !isZeroAddress(testSecondaryAddr) {
-			log.Printf("testSecondaryAddr: %s\n", testSecondaryAddr)
```
RodrigoAD, line 323: Removing noisy logs

contracts/test/Move.lock
```diff
@@ -1,29 +0,0 @@
-# Generated by move; do not edit
```
RodrigoAD, line 1: they were causing issues in CI

deployment/state_test.go
```diff
@@ -10,7 +10,6 @@ import (
 	"github.com/smartcontractkit/chainlink-deployments-framework/chain"
 	"github.com/smartcontractkit/chainlink-deployments-framework/chain/sui"
 	cldf "github.com/smartcontractkit/chainlink-deployments-framework/deployment"
-	"github.com/smartcontractkit/chainlink/v2/core/logger"
```
RodrigoAD, line 13: We don't need the logger here, also we avoid the v2 dependency this way

deployment/adapters/mcmsreader.go
```diff
@@ -0,0 +1,105 @@
...
+
+	opCount, err := inspector.GetOpCount(e.GetContext(), state.MCMSStateObjectID)
+	if err != nil {
+		return mcmstypes.ChainMetadata{}, fmt.Errorf("failed to get opCount for MCMS at %s on chain %d: %w", state.MCMSStateObjectID, chainSelector, err)
+	}
+
+	return suisdk.NewChainMetadata(
+		opCount,
+		role,
+		state.MCMSPackageID,
```
FelixFan1992, line 51: just to note that: if we use a different MCMS instance (not the regular one), we need to update these MCMS references. but for now, it's OK to keep it bc it's not decided yet.

FelixFan1992: can you add a TODO to verify this later?

FelixFan1992 (approved):
a small comment to add a TODO

PR #375, opened 2026-04-08
handle latest package id

FelixFan1992:


RodrigoAD (changes requested):
Let's please hold this one

PR #435, opened 2026-07-02
codec: create codec module

jmank88:
This PR creates a separate, top-level module `codec`, which contains the `relayer/codec` and `relayer/chainreader/loop` packages. The old packages are deprecated but remain compatible by forwarding to the new module.

Supports:
- https://github.com/smartcontractkit/chainlink/pull/23011

TODO:
- [ ] Does CI run the unit tests in their new location?
- [ ] Split into two PRs to satisfy CI check for non-default branch commit hashes


codec/loop/loop_reader_test.go
jmank88, line 1: These tests use packages from the main module which pollutes the go.mod. We can either leave the tests behind the main module only, or find a way to adapt them.

codec/types.go
FelixFan1992, line 1: types are migrated here

faisal-chainlink:
This is much cleaner, thanks for creating the PR! 

Small heads up that I might make some minor changes to the codec in the next couple of days as I complete the o11y integration.

codec/types.go
faisal-chainlink: I wonder if those types belong to the top-level codec module, not a hard opinion though, I can see it working both here and in `relayer/common`

codec/loop/loop_reader_test.go
faisal-chainlink: we can keep them in the main module since they're technically integration-ish tests (they test CR & Indexer functionality indirectly)

FelixFan1992:
> This is much cleaner, thanks for creating the PR!
> 
> Small heads up that I might make some minor changes to the codec in the next couple of days as I complete the o11y integration.

can we merge this so your change is on top of this?

faisal-chainlink (approved):

PR #455, opened 2026-07-13
fix fastcurse and mcmsreader

FelixFan1992:
this PR:
1. in fastcurse, load Sui addresses from datastore (address_refs) first. if failed, load Sui addresses from address book
2. route MCMS qualifier: UltraFastCurse, RMNMCMS, and fastcurse-devenv-curse to fast MCMS. other qualifiers will be routed to slow MCMS

### DIFF
--- a/deployment/adapters/fastcurse.go
+++ b/deployment/adapters/fastcurse.go
@@ -45,6 +45,8 @@ func NewCurseAdapter() *CurseAdapter {
 }
 
 // Initialize populates the adapter's state fields from the on-chain state for the given selector.
+// Chain metadata is resolved via LoadOnchainStatesui, which prefers datastore address refs
+// (address_refs.json) and falls back to the legacy address book (addresses.json).
 func (c *CurseAdapter) Initialize(e cldf.Environment, selector uint64) error {
 	stateMap, err := deployment.LoadOnchainStatesui(e)
 	if err != nil {
--- a/deployment/state_sources.go
+++ b/deployment/state_sources.go
@@ -1,0 +1,78 @@
+package deployment
+
+import (
+	"errors"
+	"fmt"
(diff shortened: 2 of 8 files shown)


JohnChangUK (approved):

deployment/adapters/fastcurse.go
```diff
@@ -45,6 +45,8 @@ func NewCurseAdapter() *CurseAdapter {
 }
 
 // Initialize populates the adapter's state fields from the on-chain state for the given selector.
+// Chain metadata is resolved via LoadOnchainStatesui, which prefers datastore address refs
+// (address_refs.json) and falls back to the legacy address book (addresses.json).
```
rodrigombsoares, line 49: This comment doesn't make too much sense here

deployment/state_sources.go
```diff
@@ -0,0 +1,78 @@
...
+	"strings"
+
+	fdatastore "github.com/smartcontractkit/chainlink-deployments-framework/datastore"
+	cldf "github.com/smartcontractkit/chainlink-deployments-framework/deployment"
+)
+
+// addressesForSuiChain loads address/type metadata for one Sui chain.
+// It prefers env.DataStore address refs (address_refs.json) when present and
+// falls back to env.ExistingAddresses (addresses.json).
+func addressesForSuiChain(env cldf.Environment, chainSelector uint64) (map[string]cldf.TypeAndVersion, error) {
```
rodrigombsoares, line 15: As long as a single address exists on datastore this will completely ignore address book without any warnings. And we know that datastore is not empty.
I don't know if that's the best approach. Maybe making sure that all of the `addresses.json` appear on `address_refs.json` and completely deprecate address book reads on demand.

PR #486, opened 2026-08-03
MCMS Sui entrypoint encoder unmapped callbacks

JohnChangUK:
## Summary
Fix `bindings/mcms_encoder.go`'s `CCIPEntrypointArgEncoder` fell through to a hardcoded `fee_quoter::mcms_apply_fee_token_updates` 3-arg shape for any MCMS callback it didn't explicitly handle — wrong for almost anything else (missing type args, missing `CCIPObjectRef`/`Coin<T>`/`DenyList`/`DeployerState`, or an extra unwanted argument).


RodrigoAD (approved):
Looks good, although we should find a better way to test this. We could add integration tests cases for the functions we are sure we'll use

JohnChangUK:
> Looks good, although we should find a better way to test this. We could add integration tests cases for the functions we are sure we'll use

Integration tests already exist for these and exercise them

PR #491, opened 2026-08-06
Enhancement: add relayer rescan / replay command

faisal-chainlink:
core ref: 84d0186177a9da8de4b754700989c8bcf6a0cdac

---

## Describe your changes

* Removes the `RescanRecent` indexer method previously trigged by contract binds
* Implements the plugin `Replay` method to allow for rescanning specific checkpoints. This currently works by rescanning a a set number of checkpoints starting from the checkpoint # specified in the `Replay` call params. 
    * **Important** - if the `ReplayCheckpointCount` config is set to 0. Replays will be from the specified checkpoint until the latest checkpoint. This scenario would be avoided and is only kept for flexibility. The currently default is set to 100, leaving the `ReplayCheckpointCount` value nil will use the default. Otherwise, using a small value such as 5 is sufficient.

## Issue ticket number and link

> ..

## Describe highly relevant files or code snippets that are critical in the review

> ..

## Are there other PRs that should be merged first?

> ..

### DIFF
--- a/relayer/chainreader/indexer/chain_poller.go
+++ b/relayer/chainreader/indexer/chain_poller.go
@@ -36,8 +36,9 @@ const (
 	minStalledTicksBeforeRetry = 2
 	// chunkQueueSizeMultiplier sizes the chunk queue relative to the worker count.
 	chunkQueueSizeMultiplier = 4
-	// defaultRescanRewindCheckpoints is how far RescanRecent rewinds when no explicit backfill window is set.
-	defaultRescanRewindCheckpoints uint64 = 100
+	// defaultRescanCheckpointCount is how many checkpoints RescanFrom covers when the config does
+	// not set ReplayCheckpointCount.
+	defaultRescanCheckpointCount uint64 = 100
 	// minStartSequence is a value used as a fallback to avoid using 0 and doing a full chain rescan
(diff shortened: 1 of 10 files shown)


relayer/chainreader/indexer/chain_poller.go
```diff
@@ -511,34 +522,52 @@ func (cp *ChainPoller) getLatestCheckpointSequence(ctx context.Context) (uint64,
...
+	latestCheckpoint, err := cp.getLatestCheckpointSequence(ctx)
 	if err != nil {
-		cp.logger.Errorw("Failed to get latest checkpoint", "error", err)
-		return
+		return fmt.Errorf("failed to get latest checkpoint: %w", err)
+	}
+
+	if fromSeq > latestCheckpoint {
+		return fmt.Errorf("rescan start checkpoint %d is ahead of the latest checkpoint %d", fromSeq, latestCheckpoint)
 	}
```
faisal-chainlink, line 537: Could also add a check against a the current watermark / next dispatchable chunk to avoid rescanning triggers unnecessarily. Not critical since this is now manually triggered via a CLI command.

faisal-chainlink:
core ref: 84d0186177a9da8de4b754700989c8bcf6a0cdac

FelixFan1992 (approved):

PR #493, opened 2026-08-10
NONEVM-4183: sui tooling writes to datastore

FelixFan1992:
1. dual write addresses to address book and datastore
2. migrate first few changesets to use this dual write mechanism


FelixFan1992:
core ref: 33caac5deecf9028b799b43bbefd14c475bd41fd

rodrigombsoares (approved):

